PR #2, opened 2023-05-20
Convert Tokens

aSemy:
Converting all Token classes.

depends on 
* #4 

The only significant change is that I changed `Token` to be a `sealed class`. I'm anticipating that this will be useful later on...


src/main/java/org/snakeyaml/engine/v2/tokens/AnchorToken.kt
```diff
@@ -11,30 +11,21 @@
...
-public final class AnchorToken extends Token {
-
-  private final Anchor value;
-
-  public AnchorToken(Anchor value, Optional<Mark> startMark, Optional<Mark> endMark) {
-    super(startMark, endMark);
-    Objects.requireNonNull(value);
-    this.value = value;
+  init {
+    Objects.requireNonNull(value)
```
krzema12: I think the null check is inserted by Kotlin, but we can clean up such stuff later.

krzema12 (approved):

src/main/java/org/snakeyaml/engine/v2/tokens/AnchorToken.kt
aSemy: yeah I cleaned this one up in a later commit (just don't ask me which commit...)

PR #6, opened 2023-05-20
Convert Exceptions

aSemy:
Depends on 

* #2 
* #3 
* #5 

Changes of note:

* I marked some functionality in `Mark` as `internal`, because it was only used in tests. (We could add an `@OptIn` annotation later)
* I tidied up the `MarkedYamlEngineException.toString()`


aSemy:
For fun, here's the current language stats (using Tokei)

```
> tokei -t=Java,Kotlin
===============================================================================
 Language            Files        Lines         Code     Comments       Blanks
===============================================================================
 Java                  174        21809        14741         4981         2087
 Kotlin                 78         4314         1830         2128          356
===============================================================================
 Total                 252        26123        16571         7109         2443
===============================================================================
```

There's a way to go yet!

krzema12 (approved):

PR #11, opened 2023-05-20
Convert core constructors

aSemy:
Depends on 

* #2
* #3
* #5
* #6
* #7
* #8
* #9
* #10 

I did a bit of code golfing to make the parsing more Kotlin-esque, so there were no mutable vars.


src/main/java/org/snakeyaml/engine/v2/constructor/ConstructScalar.kt
```diff
@@ -31,7 +31,7 @@ abstract class ConstructScalar : ConstructNode {
 
     companion object {
         @JvmField
-        protected val BOOL_VALUES: Map<String, Boolean> = mapOf(
+        internal val BOOL_VALUES: Map<String, Boolean> = mapOf(
```
aSemy, line 34: Kotlin doesn't allow `@JvmField` + `protected` for some obscure reason - it causes a compiler error.

krzema12 (approved):

PR #13, opened 2023-05-20
add `@Language("file-reference")` for extra clickability in test utils

aSemy:
Depends on

* No dependencies!

Minor QOL improvement in the Test DSL. Annotating the file paths with `@Language("file-reference")` means the files can be opened in IntelliJ by ctrl+clicking the file path.

Missing files are highlighted.

image

### Summary of changes

For this to work, 

* all paths need to be absolute from the root resources (all paths were already the full file path, I just needed to add the base `/`)
* use `java.getResouce()` instead of `classLoader.getResource()`.

### Impact

This is a test-only change and does not have any impact on the library.

### DIFF
--- a/src/test/java/org/snakeyaml/engine/usecases/inherited/InheritedImportTest.java
+++ b/src/test/java/org/snakeyaml/engine/usecases/inherited/InheritedImportTest.java
@@ -13,14 +13,6 @@
  */
 package org.snakeyaml.engine.usecases.inherited;
 
-import static org.junit.jupiter.api.Assertions.assertTrue;
-
-import java.io.File;
-import java.io.FilenameFilter;
-import java.io.IOException;
-import java.io.InputStream;
-import java.util.ArrayList;
-import java.util.List;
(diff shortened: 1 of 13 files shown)


src/test/java/org/snakeyaml/engine/usecases/inherited/InheritedImportTest.java
```diff
@@ -29,29 +21,39 @@
...
 
-  public static final String PATH = "inherited_yaml_1_1";
+  public static final String PATH = "/inherited_yaml_1_1";
 
 
   protected String getResource(String theName) {
-    return TestUtils.getResource(PATH + File.separator + theName);
+    //noinspection InjectedReferences
+    return TestUtils.getResource(PATH + "/" + theName);
   }
```
aSemy, line 42: IntelliJ didn't understand the path reference when it's based on a variable, so I disabled the injection for this method

krzema12 (approved):

PR #16, opened 2023-05-21
minor quick fixes in Java code, to help prep for Kotlin migration

aSemy:


src/main/java/org/snakeyaml/engine/v2/nodes/Node.kt
```diff
@@ -121,7 +121,7 @@ abstract class Node @JvmOverloads constructor(
   /**
    * @return scalar, sequence, mapping
    */
-  abstract val nodeType: NodeType?
+  abstract val nodeType: NodeType
```
aSemy, line 124: all implementations of the property are non-nullable, I think this property was marked null by mistake

PR #17, opened 2023-05-22
Convert comments

krzema12:
Convert a single package. Nothing really interesting to elaborate on.


src/main/java/org/snakeyaml/engine/v2/comments/CommentLine.kt
```diff
@@ -0,0 +1,90 @@
...
+    init {
+        Objects.requireNonNull(startMark)
+        this.startMark = startMark
+        Objects.requireNonNull(endMark)
+        this.endMark = endMark
+        Objects.requireNonNull(value)
+        this.value = value
+        Objects.requireNonNull(commentType)
+        this.commentType = commentType
+    }
```
aSemy, line 85: `requiresNonNulll()` is no longer required thanks to non-nullable types, so I propose removing them as we go

```suggestion
    init {
        this.startMark = startMark
        this.endMark = endMark
        this.value = value
        this.commentType = commentType
    }
```

aSemy (approved):

PR #20, opened 2023-05-23
Convert Composer

aSemy:
Converted Composer to Java.

I made some minor adjustments to the code to make it more Kotlin-esque.

* changes the `resolved` values to be initialised as vals, not vars
* updated the return types of Node composing functions to return the specific type, not the `Node` supertype

I noticed a couple of potential future improvements that I'll comment on to highlight, but I didn't want to make the changes here during the migration.


src/main/java/org/snakeyaml/engine/v2/composer/Composer.kt
```diff
@@ -0,0 +1,359 @@
...
+        return node
+    }
+
+    private fun composeNode(parent: Optional<Node>): Node {
+        blockCommentsCollector.collectEvents()
+        parent.ifPresent { e: Node ->
+            recursiveNodes.add(e) // TODO add unit test for this line
+        }
+        val node: Node
+        if (parser.checkEvent(Event.ID.Alias)) {
```
aSemy, line 157: This type-checking looks like a good candidate for making `Event` a sealed class and using a 'when' statement.

src/main/java/org/snakeyaml/engine/v2/composer/Composer.kt
```diff
@@ -0,0 +1,359 @@
...
+     */
+    private fun composeKeyNode(node: MappingNode): Node = composeNode(Optional.of(node))
+
+    /**
+     * To be able to override `composeNode(node)` which is a value
+     *
+     * @param node - the source
+     * @return node
+     */
+    private fun composeValueNode(node: MappingNode): Node = composeNode(Optional.of(node))
```
aSemy, line 358: I don't understand the purpose of the `composeKeyNode()` and `composeValueNode()` functions.

I suspect this code can be simplified to 

```kotlin
    private fun composeMappingChildren(children: MutableList<NodeTuple>, node: MappingNode) {
        val itemKey = composeNode(Optional.of(node))
        val itemValue = composeNode(Optional.of(node))
        children.add(NodeTuple(itemKey, itemValue))
    }
```

especially if Optional can be replaced with nullable types.

src/main/java/org/snakeyaml/engine/v2/exceptions/ComposerException.kt
```diff
@@ -24,10 +24,10 @@ import java.util.*
  * @param problemMark - the problem location
  */
 class ComposerException @JvmOverloads constructor(
-    context: String = "",
-    contextMark: Optional<Mark> = Optional.empty<Mark>(),
     problem: String?,
     problemMark: Optional<Mark>,
+    context: String = "",
+    contextMark: Optional<Mark> = Optional.empty<Mark>(),
 ) : MarkedYamlEngineException(
```
aSemy, line 31: re-ordered to put the optional parameters last, which works better during usage.

src/main/java/org/snakeyaml/engine/v2/comments/CommentEventsCollector.kt
```diff
@@ -153,11 +153,10 @@ class CommentEventsCollector {
...
-         *
-         * @return `true` if it does; Otherwise, `false`
-         */
-        get() = commentLineList.isEmpty()
+    /**
+     * Test if the collector contains any collected events.
+     *
+     * @return `true` if it does; Otherwise, `false`
+     */
+    fun isEmpty(): Boolean = commentLineList.isEmpty()
```
aSemy, line 161: renamed to match the JavaBeans naming conventions (properties shouldn't start with `is`), because I think it's nice to be consistent.

Issue #19, opened 2023-05-22
Update commit history


aSemy:
In order to clean up the commit history we could do an update when everything is converted.

We are currently committing to `master` branch.

1. make a new fork, `main`, based on the original Java code
2. in original rename all `.java` files to `.kt`, then `git commit`
3. copy the Kotlin code from `master` into `main`, then `git commit`
4. Update the project's default branch to `main`
5. (optional) delete `master` branch

That would mean the commit history is cleaner (only two commits) and also the git history would be preserved for each file.

Issue #18, opened 2023-05-22
Replace `java.util.Optional` with Kotlin Multiplatform alternative


aSemy:
SnakeYAML makes heavy use of Java Optionals. These must be replaced when the project is (eventually) converted to Kotlin Multiplatform.

There’s two ways this could go. If the project just uses Optionals as the equivalent of a nullable type, then we can just replace it with a Kotlin nullable type. Lots of code will change, so it might be a little bit tricky.

However, Optionals allow for lazy evaluation. If the project is using that, then we might need to investing creating an alternative. Which shouldn’t be too much work, but it might be complicated.

PR #27, opened 2023-05-23
Convert Scanner

### LINKED ISSUE #22: Convert Java files in `org.snakeyaml.engine.v2.scanner` to Kotlin
https://github.com/krzema12/snakeyaml-engine-kmp/tree/ddea85d48d8a91abbda7a71120a8d4281a01165c/src/main/java/org/snakeyaml/engine/v2/scanner

aSemy:
Fix #22 

Okay, now things are getting difficult, this was a lot more work! It's a large class (possibly needs some refactoring #28), so the auto-conversion couldn't cope with it all at once, I had to do it piece-by-piece. (The git history will be fixed by #19)

This wasn't a straight conversion, I also made some Kotlinizations:

* Converted Chomper to a sealed type, and added some util properties to make usage a little more plain
* Removed some `Optional<>`s (starting work on #18)
* Refactored some code to use `val`s instead of `var`s (e.g. by adding an extra `else {}` statement that sets the value to `null`)
* changed `tokens` to be an `ArrayDeque` instead of a mutable list (shouldn't have an impact, but it made more sense to me to use a explicit queue-like structure rather than an non-specific collection)


src/main/java/org/snakeyaml/engine/v2/scanner/SimpleKey.kt
```diff
@@ -0,0 +1,59 @@
...
+    @JvmName("getLineJvm")
+    fun getLine() = line
+
+    @Deprecated("help during java->kt auto convert", ReplaceWith("column"))
+    @JvmName("getColumnJvm")
+    fun getColumn() = column
+
+    @Deprecated("help during java->kt auto convert", ReplaceWith("mark"))
+    @JvmName("getMarkJvm")
+    fun getMark() = mark
```
aSemy, line 55: These sorts of functions really help IJ with the auto-convert, because IJ doesn't realise that a Java `getMark()` can be converted to `mark`.

krzema12 (approved):

PR #36, opened 2023-05-26
Convert Serializer package to Kotlin/JVM

### LINKED ISSUE #35: Convert Java files in `org.snakeyaml.engine.v2.serializer` to Kotlin
https://github.com/krzema12/snakeyaml-engine-kmp/tree/a911bd4ebc6b957a8b1c5f1c9225070ab9f247f8/src/main/java/org/snakeyaml/engine/v2/serializer

aSemy:
fix #35 

The last Java package is converted to Kotlin/JVM!

I also performed some tidying up in `DumpSettingsBuilder`


src/main/java/org/snakeyaml/engine/v2/serializer/NumberAnchorGenerator.kt
```diff
@@ -0,0 +1,38 @@
...
+import org.snakeyaml.engine.v2.common.Anchor
+import org.snakeyaml.engine.v2.nodes.Node
+
+/**
+ * Simple generate of the format id + number
+ *
+ * @param lastAnchorId - the number to start from
+ */
+class NumberAnchorGenerator(
+    private var lastAnchorId: UInt = 0u
```
aSemy, line 25: I thought it made sense to use a UInt instead of an Int here. There shouldn't be any practical reason why, but it helps clarify that the ID shouldn't be negative

krzema12 (approved):

PR #38, opened 2023-05-26
remove Optional, replace with nullable types

### LINKED ISSUE #18: Replace `java.util.Optional` with Kotlin Multiplatform alternative

aSemy:
fix #18 

Replaces all `Optional<>`s with a nullable equivalent (except for the Optional encoder/decoder).

### DIFF
--- a/src/main/java/org/snakeyaml/engine/v2/common/FlowStyle.kt
+++ b/src/main/java/org/snakeyaml/engine/v2/common/FlowStyle.kt
@@ -18,18 +18,18 @@ package org.snakeyaml.engine.v2.common
  * styles rely on explicit indicators to denote nesting and scope.
  */
 enum class FlowStyle {
-  /**
-   * Flow style
-   */
-  FLOW,
+    /**
+     * Flow style
+     */
+    FLOW,
 
-  /**
-   * Block style
-   */
-  BLOCK,
+    /**
+     * Block style
+     */
+    BLOCK,
 
-  /**
-   * Block style for the root level and flow style for other levels
-   */
-  AUTO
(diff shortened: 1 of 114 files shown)


src/main/java/org/snakeyaml/engine/v2/common/FlowStyle.kt
```diff
@@ -18,18 +18,18 @@ package org.snakeyaml.engine.v2.common
...
  */
 enum class FlowStyle {
-  /**
-   * Flow style
-   */
-  FLOW,
+    /**
+     * Flow style
+     */
+    FLOW,
```
krzema12, line 24: Looks like accidental formatting change, but I don't want you to be blocked on this. Later on we can configure e. g. ktlint.

aSemy: I think I converted this file before the EditorConfig was added, and it was never updated.

And yeah, adding a better way of formatting would be nice. I'm not a big fan of ktlint, but EditorConfig doesn't seem to work as well as I remember. At least committing the `.idea/codeStyles/` would help.

PR #44, opened 2023-05-30
Set up Kotlin Multiplatform build config

aSemy:
* Sets up Kotlin Multiplatform build config
* Only targets Kotlin/JVM and Kotlin/JS at present, but the wiring is there for Kotlin/Native
* Move JVM files into `src/jvmMain` and `src/jvmTest`
* adds a new GitHub Workflow step for caching Kotlin Konan

(there are a lot of changes but most of this is just moving files)

Depends on 

- [x] #43 

Config cache is disabled, because Kotlin 1.8 is not compatible. It can be re-enabled with Kotlin 1.9 https://github.com/krzema12/snakeyaml-engine-kmp/issues/46.


build.gradle.kts
```diff
@@ -1,46 +1,44 @@
...
 plugins {
-    kotlin("jvm") version "1.8.21"
-}
-
-dependencies {
-    testImplementation("org.junit.jupiter:junit-jupiter-engine:5.4.0")
-    testImplementation("com.google.guava:guava:26.0-jre")
+    buildsrc.conventions.lang.`kotlin-multiplatform-jvm`
+    buildsrc.conventions.lang.`kotlin-multiplatform-js`
+    //buildsrc.conventions.lang.`kotlin-multiplatform-native`
```
krzema12, line 7: It's commented out, should it stay here?

build.gradle.kts
```diff
@@ -1,46 +1,44 @@
...
         }
     }
-}
 
-tasks.withType<JavaCompile> {
-    options.encoding = "UTF-8"
+    sourceSets {
+        commonMain {
+            dependencies {
+                //implementation("com.squareup.okio:okio:3.3.0") // TODO https://github.com/krzema12/snakeyaml-engine-kmp/issues/42
```
krzema12, line 28: Commented out - shall we remove?

buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,28 @@
...
+
+plugins {
+   `kotlin-dsl`
+}
+
+dependencies {
+    implementation("org.jetbrains.kotlin:kotlin-gradle-plugin:1.8.21")
+}
+
+//val buildLogicJvmTarget = "11"
```
krzema12, line 11: Ditto.

krzema12 (approved):

buildSrc/build.gradle.kts
aSemy: Good spot, most of the comments here can be removed. I'll make another PR.

build.gradle.kts
aSemy: I'm starting to use this in the upcoming branch https://github.com/krzema12/snakeyaml-engine-kmp/tree/kmp/conversion

build.gradle.kts
aSemy: It'll be enabled eventually, but I didn't want to add _loads_ of Kotlin/Native targets right now because it will make development slower - just having two targets is enough to make sure that commonMain doesn't contain any JVM code.

PR #45, opened 2023-05-30
Convert Escapers to Kotlin Multiplatform

aSemy:
Depends on 

- [ ] #43 
- [ ] #44


aSemy:
I've found an alternative library instead https://github.com/ethauvin/urlencoder/ which is much more suitable for Kotlin Multiplatform. I'll make a new PR.

PR #48, opened 2023-06-01
Convert code to Kotlin Multiplatform

aSemy:
Converts all Kotlin/JVM code to Kotlin/Multiplatform.

There are no tests for Kotlin/JS at the moment, but it compiles!


src/commonMain/kotlin/org/snakeyaml/engine/v2/emitter/Emitter.kt
```diff
@@ -274,39 +276,31 @@ class Emitter(
...
             }
-            val chunks = java.lang.StringBuilder()
-            val start = 0
-            var end = if (prefix[0] == '!') 1 else 0
-            while (end < prefix.length) {
-                end++
-            }
-            chunks.append(prefix, start, end)
-            return chunks.toString()
         }
```
krzema12, line 304: .

PR #52, opened 2023-06-07
Enable Kotlin Native

aSemy:
* tidied up the K/N convention plugin
* removed some of the more esoteric targets (which can always be re-enabled)
* added more runners to the workflow


build.gradle.kts
```diff
@@ -6,7 +6,7 @@ plugins {
     buildsrc.conventions.lang.`kotlin-multiplatform-js`
     // Kotlin/Native is disabled for now because it will make the build slower during this phase of development.
     // Native target can be enabled when the project is more stable (after all code is KMP).
```
krzema12, line 8:
```suggestion
```

krzema12 (approved):

PR #54, opened 2023-06-07
setup git-branch-publish util

aSemy:
Adds a GH Workflow and Gradle config that will publish SnakeYAML to the (orphaned) [artifacts](https://github.com/krzema12/snakeyaml-engine-kmp/tree/artifacts) branch.

@krzema12 can you do the following:

- [x] Creat a branch protection rule for `artifacts`. It doesn't need to do anything, except exist, because then it will make the branch non-deletable.
- [x] [Create an environment](https://docs.github.com/en/actions/deployment/targeting-different-environments/using-environments-for-deployment), named 'artifacts'. (I did this for Dokkatoo, but I can't remember why...)

### DIFF
--- a/.github/workflows/git-branch-publish.yml
+++ b/.github/workflows/git-branch-publish.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: Run publish task
         run: >-
-          ./gradlew publishAllPublicationsToGitHubPublishRepository --stacktrace --info
+          ./gradlew publishAllPublicationsToGitBranchPublishRepository --stacktrace --info
         env:
           GIT_BRANCH_PUBLISH_DIR: ${{ runner.temp }}/m2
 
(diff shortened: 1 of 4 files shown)


.github/workflows/git-branch-publish.yml
```diff
@@ -0,0 +1,94 @@
+name: Git Branch Publish
```
krzema12, line 1: Are you fine with me using github-workflows-kt for this in a separate PR?

krzema12 (approved):

build.gradle.kts
```diff
@@ -5,11 +5,12 @@ plugins {
     buildsrc.conventions.lang.`kotlin-multiplatform-jvm`
     buildsrc.conventions.lang.`kotlin-multiplatform-js`
     buildsrc.conventions.lang.`kotlin-multiplatform-native`
+    buildsrc.conventions.`git-branch-publish`
 }
 
 group = "org.snakeyaml"
 version = "2.7-SNAPSHOT"
```
krzema12, line 12: Let's discuss separately how we should version the library. There's also another option, to start from 0.x.

aSemy: yes, that crossed my mind too 👍. Also, potentially an entirely new group & name?

PR #55, opened 2023-06-07
add concurrency & cancellation to 'build' workflow

aSemy:
This will cancel in-progress workflows if a new commit is added.

Adding this is especially important now that most Kotlin targets are activated, so the tests take a long time to complete.


.github/workflows/build.main.kts
```diff
@@ -19,6 +20,10 @@ workflow(
         PullRequest(),
     ),
     sourceFile = __FILE__.toPath(),
+    concurrency = Concurrency(
+        group = "\${{ github.workflow }} @ \${{ github.event.pull_request.head.label || github.head_ref || github.ref }}",
```
krzema12, line 24: You could make it more type-safe with `expr {}`, but it's a nitpick :)

krzema12 (approved):

.github/workflows/build.main.kts
aSemy: This was a quick copy & paste, so feel free to update it with `expr {}` 👍

PR #61, opened 2023-06-11
fix error in JS when overriding exception message

aSemy:


src/commonMain/kotlin/org/snakeyaml/engine/v2/exceptions/MarkedYamlEngineException.kt
```diff
@@ -17,48 +17,75 @@ package org.snakeyaml.engine.v2.exceptions
...
- * @param problemMark - position of the issue
  */
-open class MarkedYamlEngineException protected constructor(
-    val context: String?,
-    val contextMark: Mark?,
-    val problem: String?,
-    val problemMark: Mark?,
+open class MarkedYamlEngineException private constructor(
+    // can't override `message` property, so define a new one
+    // https://youtrack.jetbrains.com/issue/KT-52653/
```
krzema12, line 23: This one seems to be the original tracking issue, and it's also good to see that it's goinf to be fixed in 1.9.0.
```suggestion
    // https://youtrack.jetbrains.com/issue/KT-43490/
```

krzema12 (approved):

PR #64, opened 2023-06-12
Commit IJ codestyle files

aSemy:
IntelliJ does not seem to support EditorConfig very well, and keeps ignoring the added settings. Committing the `.idea/codeStyle/` files will improve things.

I have committed the IJ code style files, with some amendments

* The Java style has been set to use a 2-space indent, to match the existing SnakeYAML code
* I have enabled vertical-alignment of `when` case branches, because I think it's prettier
* star-importing of `org.gradle.kotlin.dsl`, which is a default-import in `.gradle.kts` files but not in `.kt` files, so adding it helps when writing Gradle utilities in `.kt` files

Additional changes:

* Tidy `.gitignore`
* tidy `.editorconfig`


.editorconfig
```diff
@@ -1,14 +1,14 @@
...
 end_of_line = lf
-indent_size = 4
 indent_style = space
 insert_final_newline = true
 trim_trailing_whitespace = true
-ij_kotlin_allow_trailing_comma = true
-ij_kotlin_allow_trailing_comma_on_call_site = true
-ij_kotlin_packages_to_use_import_on_demand = unset
-ij_kotlin_name_count_to_use_star_import = 2147483647
-ij_kotlin_name_count_to_use_star_import_for_members = 2147483647
```
aSemy, line 14: Perhaps controversially I have not retained the 'no-wildcard-imports' config. I don't think the style-guide has anything to say on these, and the IJ default is to allow them, so they are allowed. 

I personally like wildcard imports, despite originally being against them. Perhaps this is because Kotlin generally handles imports and type-inference more stably than Java, so it's less of an issue? Or because Kotlin libraries often have extension functions that are much more convenient to use when wildcard-importing?

In any case, are you okay with this decision?

krzema12: I personally like to know where my imports come from, but I won't oppose here - it's a minor issue to me :)

krzema12 (approved):

PR #65, opened 2023-06-12
fix result of `flattenMapping()` being ignored

aSemy:
This should have been included in #60


src/commonMain/kotlin/org/snakeyaml/engine/v2/constructor/StandardConstructor.kt
```diff
@@ -114,13 +114,13 @@ open class StandardConstructor(settings: LoadSettings) : BaseConstructor(setting
...
 
     override fun constructMapping2ndStep(node: MappingNode, mapping: MutableMap<Any?, Any?>) {
-        flattenMapping(node)
-        super.constructMapping2ndStep(node, mapping)
+        val flattened = flattenMapping(node)
+        super.constructMapping2ndStep(flattened, mapping)
     }
 
     override fun constructSet2ndStep(node: MappingNode, set: MutableSet<Any?>) {
-        flattenMapping(node)
```
krzema12, line 122: Weird there's no need to adjust unit tests :thinking: is it possible to cover it?

krzema12: If you think some higher-level tests should cover it, let's go ahead with what we have now!

krzema12 (approved):

src/commonMain/kotlin/org/snakeyaml/engine/v2/constructor/StandardConstructor.kt
aSemy: Hmm, I'm not sure there are any tests. I've had a play around with trying to adapt an existing test, but I couldn't create a failing test.

@asomov, maybe you could help try and create a test for this scenario? I suspect SYE2 is also missing coverage for this scenario, do you know if that's the case?

asomov: @aSemy I do not get this PR. 
flattenMapping(node) changes the node in place.

aSemy: We changed `MappingNode.value` to be immutable in #60, so `flattenMapping(node)` returns a new instance if there's a change. I forgot to update `constructMapping2ndStep()` and `constructSet2ndStep()` in #60 though, so this PR is rectifying that. 

What we're puzzled about is that there wasn't a test to cover this situation. My guess is that in SYE `flattenMapping(node)` can be commented out and no test will fail.

asomov: You can to comment it out and check ! Do you have all the tests ?

aSemy: Good point! I've checked and while commenting out `flattenMapping(node)` causes errors, if I comment out 


```java
nodeValue.remove(indices2remove.next().intValue());
```

https://github.com/snakeyaml/snakeyaml-engine/blob/5a35e1fe7f780d1405d5a03470f9f13d32b1638a/src/main/java/org/snakeyaml/engine/v2/constructor/StandardConstructor.java#L97-L100

then no tests fail. So `flattenMapping(node)` is only used to verify that there are no duplicates, and it doesn't make a difference if `flattenMapping(node)` changes the node in place or not.

asomov: Yes, the javadoc:
   _* Flattening is not required because merge was removed from YAML 1.2 Only check duplications_

aSemy: Riiight okay, that's what that meant, thanks for explaining. 

So the mutation code in `StandardConstructor.processDuplicateKeys(MappingNode node)` is intended to have no effect. Should it be removed?

asomov: you can try to remove and observe the result

aSemy: Done, and the tests pass, thanks for your insight @asomov 🙏 

I've refactored the code to make it more clear that flattening isn't required any more, and that the duplicate-key validation doesn't need to mutate the node.

asomov: I do not get why this test does not fail for you:
```
  @Test
  @DisplayName("Do not allow duplicate keys")
  void doNotAllowDuplicateKeys() {
    LoadSettings settings = LoadSettings.builder().setAllowDuplicateKeys(false).build();
    Load load = new Load(settings);
    try {
      load.loadFromString("{a: 1, a: 2}");
      fail("Duplicate keys must not be allowed.");
    } catch (DuplicateKeyException e) {
      assertTrue(e.getMessage().contains("found duplicate key a"));
    }
  }
```

aSemy: Duplicate keys are detected in [`validateDuplicateKeys(node)`](https://github.com/krzema12/snakeyaml-engine-kmp/blob/3e229811c580e1a682a5252d45203e7e67c6469b/src/commonMain/kotlin/org/snakeyaml/engine/v2/constructor/StandardConstructor.kt#L44-L65), which is called in [`constructMapping2ndStep()`](https://github.com/krzema12/snakeyaml-engine-kmp/blob/3e229811c580e1a682a5252d45203e7e67c6469b/src/commonMain/kotlin/org/snakeyaml/engine/v2/constructor/StandardConstructor.kt#L87-L91).

Issue #85, opened 2023-09-09
Port and release snakeyaml-engine 2.7


krzema12:
Last commit from the parent repo: https://github.com/krzema12/snakeyaml-engine-kmp/commit/04614c75c124d30c58221ab11255c5d45d913a14
Commits to port for releasing 2.7: https://bitbucket.org/snakeyaml/snakeyaml-engine/branches/compare/snakeyaml-engine-2.7%0D04614c7#commits

We should release snakeyaml-engine-kmp as 2.7.1 to reflect snakeyaml-engine 2.7. Later on, we'll track the parent project's version.

krzema12:
https://github.com/krzema12/snakeyaml-engine-kmp/releases/tag/v2.7.1

PR #90, opened 2023-09-12
[#85] Sync with parent project's tag snakeyaml-engine-2.7

krzema12:
https://bitbucket.org/snakeyaml/snakeyaml-engine/branches/compare/snakeyaml-engine-2.7%0D04614c7#diff

It's a single commit because there are several merge commits along the way,
and recreating the exact commit structure would be hard.


aSemy (approved):
Good job!

PR #115, opened 2023-12-17
tidy build config after Gradle & Kotlin updates

aSemy:
- enable Gradle config cache (Kotlin 1.9 is now compatible)
- update Gradle config cache property names
- use Kotlin default hierarchy https://kotlinlang.org/docs/multiplatform-hierarchy.html#default-hierarchy-template
- fix Kotlin config deprecations

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -23,7 +23,7 @@ kotlin {
             kotlin.srcDirs(tasks.yamlTestSuiteDataSources)
 
             dependencies {
-                implementation(platform("io.kotest:kotest-bom:5.8.0"))
+                implementation(project.dependencies.platform("io.kotest:kotest-bom:5.8.0"))
                 implementation("io.kotest:kotest-framework-engine")
                 implementation("io.kotest:kotest-framework-api")
                 implementation("io.kotest:kotest-assertions-core")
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -23,7 +23,7 @@ kotlin {
             kotlin.srcDirs(tasks.yamlTestSuiteDataSources)
 
             dependencies {
-                implementation(platform("io.kotest:kotest-bom:5.8.0"))
+                implementation(project.dependencies.platform("io.kotest:kotest-bom:5.8.0"))
```
krzema12, line 26: Why this change? The former was more compact.

buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform-native.gradle.kts
```diff
@@ -51,52 +24,4 @@ kotlin {
     //iosSimulatorArm64()
     //tvosSimulatorArm64()
     //watchosSimulatorArm64()
-
-    @Suppress("UNUSED_VARIABLE")
```
krzema12, line 55: :tada:

build.gradle.kts
aSemy: It was deprecated

https://github.com/JetBrains/kotlin/blob/v1.9.21/libraries/tools/kotlin-gradle-plugin-api/src/common/kotlin/org/jetbrains/kotlin/gradle/plugin/HasKotlinDependencies.kt#L76-L81

krzema12 (approved):

Issue #124, opened 2024-01-19
Add usage sections in readme


Zane-XY:
Hi could you please add a section in readme on how to serialize and deserialize in Kotlin? I really want to have a try.

aSemy:
Hi @Zane-XY 👋 


This information should be added to the README:

snakeyaml-engine-kmp is quite low-level. You can take a look at an example usage in an end-to-end test https://github.com/krzema12/snakeyaml-engine-kmp/blob/v2.7.1/src/commonTest/kotlin/org/snakeyaml/engine/SimpleEndToEndTest.kt

If you want to encode/decode YAML with Kotlin, then it's best to use https://github.com/charleskorn/kaml. snakeyaml-engine-kmp was converted from Java specifically for Kaml.

krzema12:
There's actually some higher-level API as well, see the example:

https://github.com/typesafegithub/github-actions-typing-catalog/blob/1f94a9e94fb4bb38dffe42f217fe1735161af624/.github/workflows/update-metadata.main.kts#L77

I haven't tried parsing yet.

I'll try to come up with some quick examples for both APIs so that we can inline them in the README for convenience.

Zane-XY:
Thanks @aSemy , @krzema12 .

PR #92, opened 2023-09-13
use UrlEncoder dependency

aSemy:
### Summary

Migrate UrlEncoder util to the now converted & published UrlEncoder library https://github.com/ethauvin/urlencoder/issues/1

PR #125, opened 2024-01-20
Update README

### LINKED ISSUE #118: Adjust README to point to Maven coords

krzema12:
Fixes #118 and #124.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,8 +19,9 @@ SnakeYAML Engine KMP is based on
 ## Status
 
 The library has been successfully ported to KMP, and is consumed by [kaml](https://github.com/charleskorn/kaml),
-a library that integrates with [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization). However,
-it hasn't been battle-tested well enough yet, so please keep it in mind when integrating with critical workflows.
+a library that integrates with [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization).
+
+SnakeYAML Engine KMP uses comprehensive [YAML test suites](https://github.com/yaml/yaml-test-suite) that validate correctness. However, performance has not been evaluated.
 
 ## How to use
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -18,11 +18,29 @@ SnakeYAML Engine KMP is based on
 
 ## Status
 
-⚠️ SnakeYAML Engine KMP is **under development**.
+The library has been successfully ported to KMP, and is consumed by [kaml](https://github.com/charleskorn/kaml),
+a library that integrates with [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization). However,
+it hasn't been battle-tested well enough yet, so please keep it in mind when integrating with critical workflows.
```
aSemy, line 23:
```suggestion
The library has been successfully ported to KMP, and is consumed by [kaml](https://github.com/charleskorn/kaml),
a library that integrates with [kotlinx.serialization](https://github.com/Kotlin/kotlinx.serialization).
```

snake-kmp is using the YAML test suite, and passes as many tests as snake-java (keeping in mind it's rare for a YAML lib to pass all test-suite cases!) And the tests pass on all Kotlin targets, which I think is a big achievement!

Did you want to communicate something else by "it hasn't been battle-tested"? Could be phrased differently? Otherwise, I propose removing it.

krzema12: I actually didn't think about correctness when writing this (tests do prove what you wrote), but rather various other aspects like performance. We had to replace JDK-specific parts with KMP counterparts, and just this can result in some non-obvious changes with regard to snakeyaml-engine. E. g. See #92, there's some performance regression when switching to the third-party library - some issues may be already present in the currently released version.

Maybe let's write that this KMP port hasn't been widely adopted yet, and is relatively young? Technically we aren't sure about the level of adoption in private projects, but what I want to communicate here is that the maturity of our lib is not equivalent to snakeyaml-engine's. Do you see my point? 😄

aSemy: Yes I think that's a good point regarding performance, I've updated my suggestion - wdyt?

aSemy (approved):

PR #127, opened 2024-01-21
fix test parseManyAliasesForCollections

aSemy:
I think the test is failing on CI because the test is performing worse than expected.

[Example failure](https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/7593971736/job/20685006620#step:5:284): 

```
ReferencesTest[jvm] > Parsing with aliases may take a lot of time, CPU and memory[jvm] FAILED
    org.opentest4j.AssertionFailedError: Time was 11.039 seconds. ==> expected: <true> but was: <false>
        at app//org.junit.jupiter.api.AssertionFailureBuilder.build(AssertionFailureBuilder.java:151)
        at app//org.junit.jupiter.api.AssertionFailureBuilder.buildAndThrow(AssertionFailureBuilder.java:132)
        at app//org.junit.jupiter.api.AssertTrue.failNotTrue(AssertTrue.java:63)
        at app//org.junit.jupiter.api.AssertTrue.assertTrue(AssertTrue.java:36)
        at app//org.junit.jupiter.api.Assertions.assertTrue(Assertions.java:214)
        at app//org.snakeyaml.engine.usecases.references.ReferencesTest.parseManyAliasesForCollections(ReferencesTest.java:113)
```

* Remove the num-of-cores logic regarding minDuration - 400ms seems fine.
* Increase max duration to 15 seconds - the exact timeout doesn't particularly matter, this test is always going to be slow.
* replace manual time measurement with Java 8 time utils
* add a JUnit timeout assertion

### DIFF
--- a/src/jvmTest/java/org/snakeyaml/engine/usecases/references/ReferencesTest.java
+++ b/src/jvmTest/java/org/snakeyaml/engine/usecases/references/ReferencesTest.java
@@ -13,11 +13,6 @@
  */
 package org.snakeyaml.engine.usecases.references;
 
-import static org.junit.jupiter.api.Assertions.assertEquals;
-import static org.junit.jupiter.api.Assertions.assertTrue;
-import static org.junit.jupiter.api.Assertions.fail;
-
-import java.util.LinkedHashMap;
 import org.junit.jupiter.api.DisplayName;
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.Test;
@@ -25,6 +20,12 @@ import org.snakeyaml.engine.v2.api.Dump;
 import org.snakeyaml.engine.v2.api.DumpSettings;
 import org.snakeyaml.engine.v2.api.Load;
 import org.snakeyaml.engine.v2.api.LoadSettings;
+
+import java.time.Duration;
+import java.time.Instant;
+import java.util.LinkedHashMap;
+
+import static org.junit.jupiter.api.Assertions.*;
 
 @Tag("fast")
 public class ReferencesTest {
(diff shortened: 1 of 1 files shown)


src/jvmTest/java/org/snakeyaml/engine/usecases/references/ReferencesTest.java
```diff
@@ -95,22 +96,23 @@ public void referencesWithRecursiveKeysNotAllowedByDefault() {
   @Test
   @DisplayName("Parsing with aliases may take a lot of time, CPU and memory")
   public void parseManyAliasesForCollections() {
-    String output = createDump(25);
-    // Load
-    long time1 = System.currentTimeMillis();
-    LoadSettings settings =
+    final var minDuration = Duration.ofMillis(400);
```
krzema12, line 99: It looks very odd to me that there's a test asserting on minimum running time. I understand the upper bound, it prevents performance regressions. WDYT about leaving just the upper bound?

aSemy: Good point. I think it makes sense to verify that the test at least does something, since the result of `load.loadFromString(output)` isn't verified.

Perhaps the result can be verified - I'll have a quick look.

Aside from that, I'd prefer to keep the JVM tests as close as possible to the original SnakeJava tests, so it's easier to get updates from upstream.

krzema12: Got it. Yeah, I also noticed the result isn't verified. I also get the argument of staying close to the upstream. Let me approve and let you decide if you want to iterate on it further.

krzema12 (approved):

src/jvmTest/java/org/snakeyaml/engine/usecases/references/ReferencesTest.java
aSemy: The result is a recursive object that would be difficult to assert on! Trying to print it to console results in `java.lang.OutOfMemoryError: Java heap space`.

PR #141, opened 2024-03-20
Enable more native targets

krzema12:
The main motivation for this is pushing https://github.com/charleskorn/kaml/issues/232 further. I'd like this library to be truly multiplatform, and work with all targets supported by Kotlin.


buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform-native.gradle.kts
```diff
@@ -14,14 +14,11 @@ kotlin {
     macosX64()
     macosArm64()
 
-    // More specialised targets are disabled for now, since I anticipate low demand.
-    // They can be re-enabled, if there is demand for them.
-    // https://kotlinlang.org/docs/multiplatform-share-on-platforms.html#use-target-shortcuts
-    //ios()     // iosArm64, iosX64
-    //watchos() // watchosArm32, watchosArm64, watchosX64
-    //tvos()    // tvosArm64, tvosX64
+    ios()     // iosArm64, iosX64
```
krzema12, line 17: @aSemy there are some errors after uncommenting these: https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/8355216197/job/22869886829

I kinda understand them, but I wonder if it's possible to easily say "build for all possible targets". I want this library to be truly multiplatform, so that it's compatible with whatever platform provided by KMP someone tries to use it. Do you know if it's possible, or we need to explicitly list the targets?

aSemy: > Do you know if it's possible, or we need to explicitly list the targets?

In practice, no, it's not possible. 

I guess theoretically we could use reflection to get all possible targets, but I think that's more trouble than it's worth.

aSemy (approved):
Great!

PR #144, opened 2024-03-21
Create API reference documentation site.

aSemy:
- Use Dokkatoo to generate API reference documentation.
- Fix links to the YAML spec.
- Fix links in KDoc (`@see [...]()` doesn't work for URLs).
- Create GitHub Workflow for uploading the reference docs to GitHub Pages.

To preview the docs, checkout the branch and run `./gradlew dokGen`, and the site is generated into `./build/dokka/html/`


.github/workflows/publish-site.yml
```diff
@@ -0,0 +1,65 @@
+name: Publish Site
+
+
+on:
+  workflow_dispatch:
```
krzema12, line 5: Could we deploy on each push of a tag (=release), to not have to think about it?

README.md
```diff
@@ -23,13 +23,22 @@ SnakeYAML Engine KMP is based on
...
 
-SnakeYAML Engine KMP uses comprehensive [YAML test suites](https://github.com/yaml/yaml-test-suite) that validate correctness. However, performance has not been evaluated.
+SnakeYAML Engine KMP uses comprehensive [YAML test suites](https://github.com/yaml/yaml-test-suite)
+that validate correctness. However, performance has not been evaluated.
 
 ## How to use
 
-The library is published to Maven Central. Use the following Maven coords:
+> [!NOTE]
+> SnakeYAML Engine KMP is a low-level library, only intended to be used by libraries.
```
krzema12, line 32: It also has some higher-level API. I'd omit this change here. We already mention kaml earlier.

build.gradle.kts
```diff
@@ -5,6 +5,7 @@ plugins {
     buildsrc.conventions.publishing
     buildsrc.conventions.`git-branch-publish`
     buildsrc.conventions.`yaml-testing`
+    id("dev.adamko.dokkatoo-html") version "2.2.0"
```
krzema12, line 8: Wasn't Dokkatoo merged into Dokka?

.github/workflows/publish-site.yml
aSemy: Good idea, I'll add it, although I was expecting you to convert it to workflows-kt :)

(I copied this workflow from another project, and it was triggered by another workflow.)

build.gradle.kts
aSemy: Yes, but only as a formality, it hasn't been worked in. I'm still developing Dokkatoo in a separate repo.

README.md
aSemy: Yeah, I see your point.

My intention was to add a note to guide the majority of devs towards a more straightforward solution, because, while it's true that SnakeKMP _can_ be used at a high level, it does requires more set up.

What about this instead? Something that's more of a comparative, just providing some information for people to use.

> [!TIP]
> Are you wondering if SnakeYAML Engine KMP is for you?
>
> SnakeYAML Engine KMP offers fine-grained control and advanced functionality, which is a good fit if you need more customization and flexibility.
> 
> If you're for a hassle-free YAML parsing experience, we recommend using kaml, which is designed for simplicity and ease of integration.

.github/workflows/publish-site.yml
krzema12: Ok, I can convert it in a follow-up PR :)

README.md
krzema12: Much better!

README.md
```diff
@@ -18,32 +18,49 @@ and supports
 SnakeYAML Engine KMP is based on
 [SnakeYAML Engine](https://bitbucket.org/snakeyaml/snakeyaml-engine/).
 
+> [!TIP]
+> Are you wondering if SnakeYAML Engine KMP is for you?
+>
+> SnakeYAML Engine KMP offers fine-grained control and advanced functionality,
+> which is a good fit if you need more customization and flexibility.
+>
+> If you're for a hassle-free YAML parsing experience, we recommend using
```
krzema12, line 27:
```suggestion
> If you're for a hassle-free YAML (de)serialization experience, we recommend using
```
since both solutions can also emit YAML.

krzema12 (commented):
Just one nitpick.

krzema12 (approved):

PR #154, opened 2024-04-01
Fix setting width greater than the default

### LINKED ISSUE #152: Max width doesn't work (works in upstream)
I want to migrate from snakeyaml-engine to snakeyaml-engine-kmp (https://github.com/typesafegithub/github-workflows-kt/pull/1338). I have this custom dumper config here:
https://github.com/typesafegithub/github-workflows-kt/blob/d2b3748edb5263ed142a91c4196aca7c922dedca/github-workflows-kt/src/main/kotlin/io/github/typesafegithub/workflows/yaml/ObjectToYaml.kt#L25

It seems to be not respected in snakeyaml-engine-kmp. It was caught by a unit test:

image

I didn't have a chance to dig deeper into this yet, but it looks like a regression during porting to Kotlin.

krzema12:
Resolves #152.

It looks like a coding error when porting snakeyaml-engine to Kotlin. The original project contains:
https://github.com/krzema12/snakeyaml-engine-kmp/blob/04614c75c124d30c58221ab11255c5d45d913a14/src/main/java/org/snakeyaml/engine/v2/emitter/Emitter.java#L213-L216

which I'm reimplementing in this PR.

BTW, there's no unit test for this behavior in snakeyaml-engine, and ideally we'd like to add it there as well. It would have prevented this regression.


aSemy (approved):
👍

PR #161, opened 2024-05-05
miscellaneous refactorings

aSemy:
Refactorings that happen to fix a few test cases

- add some prep for K2 (add `.kotlin` to .gitignore, simplify K/N repos)
- A bunch of refactors to ScannerImpl
- Minor edit to 'cannot start any token' exception message (remove quote around char, don't repeat TAB)
- Add more utils to StreamReader (helps simplify ScannerImpl)
- Simplify `CharConstants.Companion.escapeChar()`
- Tidy up Kdoc
- Update JsonSchema, so it's not an expect/actual class (it's not necessary)

### DIFF
--- a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ScannerImpl.kt
+++ b/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ScannerImpl.kt
@@ -1,8 +1,5 @@
 package it.krzeminski.snakeyaml.engine.kmp.scanner
 
-import okio.Buffer
-import it.krzeminski.snakeyaml.engine.kmp.internal.utils.Character
-import it.krzeminski.snakeyaml.engine.kmp.internal.utils.appendCodePoint
 import it.krzeminski.snakeyaml.engine.kmp.api.LoadSettings
 import it.krzeminski.snakeyaml.engine.kmp.comments.CommentType
 import it.krzeminski.snakeyaml.engine.kmp.common.Anchor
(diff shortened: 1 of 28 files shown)


src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ScannerImpl.kt
```diff
@@ -1959,48 +1934,36 @@ class ScannerImpl(
...
-                        }
-                    } else {
-                        break
-                    }
+        }
+
+        val breaks = StringBuilder()
+
+        while (true) {
+            reader.forwardWhile { c -> c in " \t" }
```
aSemy, line 1952: Changing 'forward while char is a space' to 'forward while char is a space _or tab_' fixed several test suite tests.

krzema12:
Thanks for this! However, let's split and maybe omit some changes for now. Thoughts:
* "Refactorings that happen to fix a few test cases" - technically a refactoring shouldn't change the behavior, but let's omit this word for now. I think fixing several test cases would make a great separate PR (one or more if makes sense?), without any accidental fixes
* I'd hold with any preparations for K2 until we actually use it, then the changes go together with the version bump which is more logical and easier to analyze later
* I really appreciate any code simplifications, but again a separate PR with such true refactoring would make most sense to me

To explain how I approach dividing changes into PRs, I think this way: each PR is like an entry in the release notes. If users see "miscellaneous refactorings", they wouldn't know what to expect or they'd think "ah, that's just refactoring, it doesn't change how the library works". The division I suggested above more or less lets us map the PRs to the release notes entries. Of course, we can discuss further.

krzema12:
@aSemy let me close it because it would require quite a lot of work to sync it with the newest state of the code base.  As far as I see, there's nothing critical here. Always open to improvements, best if in small, isolated PRs!

PR #187, opened 2024-05-17
Kotlin 2.0.0

aSemy:
- remove KT-51379 workaround
- add `.kotlin/` to `.gitignore`
- Disable Yarn (K2 can use NPM instead - disabling Yarn is nice because it's one less tool to worry about)


aSemy:
I've experimented briefly, but I think IntelliJ doesn't support K2 very well. I kept seeing IDE errors about `kotlin.Unit` not being found.

krzema12 (approved):

krzema12:
> I've experimented briefly, but I think IntelliJ doesn't support K2 very well. I kept seeing IDE errors about `kotlin.Unit` not being found.

Do you know if this issue is tracked somewhere?

aSemy:
> Do you know if this issue is tracked somewhere? As you spend more time in the code base, so you find it a blocker for merging this PR? I use Kotlin 2.0 in github-workflows-kt and haven't noticed this problem.

So K2 works - but IntelliJ's K2 mode doesn't work for KMP projects https://blog.jetbrains.com/idea/2024/03/k2-kotlin-mode-alpha-in-intellij-idea/. So, just don't enable K2 in IntelliJ for now. I couldn't find a YouTrack issue...

aSemy:
Here's a ticket for K2 KMP support in IJ https://youtrack.jetbrains.com/issue/KTIJP-6

PR #190, opened 2024-05-22
Remove deprecated Tag constructor

aSemy:
Also, tidy up some KDoc

### DIFF
--- a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/nodes/Tag.kt
+++ b/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/nodes/Tag.kt
@@ -15,7 +15,6 @@ package it.krzeminski.snakeyaml.engine.kmp.nodes
 
 import it.krzeminski.snakeyaml.engine.kmp.common.UriEncoder
 import kotlin.jvm.JvmField
-import kotlin.reflect.KClass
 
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/nodes/Tag.kt
```diff
@@ -28,18 +27,10 @@ class Tag(
...
     val value: String = UriEncoder.encode(tag)
 
-    /**
-     * Create a global tag to dump the fully qualified class name
-     *
-     * @param clazz - the class to use the name
-     */
-    @Deprecated(
-        "Kotlin Reflection is not well supported on non-JVM platforms - manually specify the FQN instead",
-        ReplaceWith("Tag.forClass(clazz.qualifiedName)"),
```
krzema12, line 38: Is there a method `forClass`? I see only `forType`, was it a mistake? Still, this change should be fine since we're planning to bump the major version, only this `ReplaceWith` was a bit unfortunate.

aSemy: Oops yeah, looks like there was never a `forClass()` util function!

krzema12 (approved):

PR #192, opened 2024-05-24
Add NPM `package-lock.json`

aSemy:
The Yarn package lock file was removed in https://github.com/krzema12/snakeyaml-engine-kmp/pull/187, but I didn't add the new NPM `package-lock.json`, but it is required.

Re-add the 'relocate lockfile dir' logic.

Enable reportNewPackageLock so KGP creates an error if the lockfile is missing (otherwise it generates a new one and doesn't complain, which obscures the fact that one is missing!)


build.gradle.kts
```diff
@@ -89,3 +93,41 @@ dokkatoo {
         }
     }
 }
+
+tasks.withType<LockCopyTask>().configureEach {
```
krzema12, line 97: Wondering if we haven't reach the tipping point where we just pragmatically use the default yarn.

aSemy:
Well... it looks like this workaround didn't work. There's a failure in #191 

https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/9228285746/job/25392037275?pr=191#step:5:489

The diff view is very useful though!

```diff
> Task :kotlinStorePackageLock FAILED
--- D:/a/snakeyaml-engine-kmp/snakeyaml-engine-kmp/build/js/package-lock.json
+++ D:/a/snakeyaml-engine-kmp/snakeyaml-engine-kmp/gradle/kotlin-js-store/package-lock.json
@@ -10,8 +10,8 @@
       "workspaces": [
         "packages/snakeyaml-engine-kmp",
         "packages/snakeyaml-engine-kmp-test",
-        "packages\\snakeyaml-engine-kmp-wasm-js",
-        "packages\\snakeyaml-engine-kmp-wasm-js-test",
+        "packages/snakeyaml-engine-kmp-wasm-js",
+        "packages/snakeyaml-engine-kmp-wasm-js-test",
         "packages_imported/kotlin-test-js-runner/0.0.1"
       ],
       "devDependencies": {}
```

krzema12:
@aSemy let's try with yarn then?

PR #203, opened 2024-06-23
Set JVM target version as 1.8

### LINKED ISSUE #202: Build using Java8?
I found that the dependency was built using Java11 when using Kaml, which makes it impossible for me to run apps that rely on this library on Java8

`java.lang.UnsupportedClassVersionError: it/krzeminski/snakeyaml/engine/kmp/exceptions/MarkedYamlEngineException has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0`

krzema12:
Fixes #202.


aSemy:
It should be possible to use JDK21 to run Gradle and compile the code (it doesn't have to be 21, but higher is better, because newer Java versions are faster and have more fixes).

Next, set the Java release flag to 8, so that JDK21 will compile the code to Java 8 https://jakewharton.com/kotlins-jdk-release-compatibility-flag/

However, this might require removing the Java Toolchain config - see https://jakewharton.com/gradle-toolchains-are-rarely-a-good-idea/

See also: https://blog.alllex.me/posts/2023-11-03-liberal-library-tooling/

For tests, it's possible to configure them to use JDK8 as a launcher https://docs.gradle.org/8.8/userguide/toolchains.html#toolchains_for_tasks

aSemy:
However, Java 8 is ancient, so I don't feel very inclined to support it.

krzema12:
> For tests, it's possible to configure them to use JDK8 as a launcher https://docs.gradle.org/8.8/userguide/toolchains.html#toolchains_for_tasks

I want to start with this, to make it test-driven, but I'm having problems. Asked in https://kotlinlang.slack.com/archives/C19FD9681/p1719233443825339

krzema12:
Try https://github.com/square/okhttp/blob/ad4c4f02860565e7217f23713f22c215f3bfce75/build.gradle.kts#L146

krzema12:
Current problem: Java tests don't compile because they use features not present in 1.8.

src/jvmTest/java/org/snakeyaml/engine/usecases/references/ReferencesTest.java
```diff
@@ -97,17 +97,17 @@ public void referencesWithRecursiveKeysNotAllowedByDefault() {
   @Test
   @DisplayName("Parsing with aliases may take a lot of time, CPU and memory")
   public void parseManyAliasesForCollections() {
-    final var minDuration = Duration.ofMillis(400);
-    final var maxDuration = Duration.ofSeconds(15);
-    final var timeElapsed = assertTimeout(maxDuration, () -> {
+    final Duration minDuration = Duration.ofMillis(400);
```
krzema12, line 100: Setting the target version to 1.8 forced me to make the tests use 1.8-compatible constructs. I'm almost sure there's a better way to do it, like allowing higher version just for tests, but since I don't have much time before my vacation, I decided to use explicit typing for these several `var`s. I guess we don't have to rush it that much, so I can revisit once I'm back, also open to suggestions.

OptimumCode: I think you should be able to configure a higher compatibility version for tests by configuring the compilation tasks

```kotlin
tasks.named<JavaCompile>("compileTestJava") {
    sourceCompatibility = "11"
    targetCompatibility = "11"
}

tasks.named<KotlinCompile>("compileTestKotlinJvm") {
    compilerOptions {
        jvmTarget.set(JvmTarget.JVM_11)
    }
}
```

aSemy: We could update the tests to use Java 11, but we need to be able to test that the code works with Java 8 :)

`var` just isn't available in Java 8, so there's no alternative. Apart from converting the tests to Kotlin!

aSemy:
I've updated the config a bit, but we have a dependency on https://github.com/ethauvin/urlencoder, which requires Java 11.

https://github.com/ethauvin/urlencoder/blob/dd3d500496b8768a89ff94d1cd0343e739a72940/buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform-base.gradle.kts#L46-L50

krzema12:
I've asked the owner of our dependency to support Java 8: https://github.com/ethauvin/urlencoder/issues/18

Issue #204, opened 2024-06-26
Parsing OpenAI takes 57 seconds


nomisRev:
Hi all,

Thank you all for the amazing work on both snakeyaml-engine-kmp, and integrating it into KAML 👏 👏 
I was testing KAML for parsing OpenAI's OpenAPI yaml file, and this took 57 seconds compared to ~110 milliseconds using SnakeYAML.

Not entire sure if it's due to `allowAnchorsAndAliases`, because other YAML's seem to work quite fast. I've not done extensive testing, or performance testing.

krzema12:
Hi @nomisRev!

Thanks for reporting this. Do you have an example YAML file where it reproduces?

krzema12:
Ok, I've reread what you wrote and understood it's about https://github.com/openai/openai-openapi/blob/master/openapi.yaml

krzema12:
@nomisRev actually, can I ask you to prepare a minimal reproducer? Since you use this lib through kaml, I'm wondering if you're trying to map the YAML into a structure of `@Serialized` classes, or in another way. A simple reproducer would disambiguate it.

aSemy:
SnakeKMP required some changes in Kaml, so it's possible we introduced a bug on the Kaml side. It'd be good to find out if Kaml is slow, or SnakeKMP.

nomisRev:
I am going to create a reproducer with the information I got here 👍 I'll report back soon.

@krzema12 No, I was deserialising into `YamlNode` (for now).

nomisRev:
I was not entirely sure how to reproduce it without KAML. Seems all the methods are quite low-level 😅 Didn't have time to get that deep. https://github.com/nomisRev/snakeyaml-kmp-reproducer

> OpenAI and OpenAPI look alike when reading too fast!

Haha, it has caused me quite a lot of confusion too!

krzema12:
Thanks @nomisRev!

From my initial tests, something interesting emerges: the regression happened in kaml 0.60.0, not in 0.58.0 where we switched to snakeyaml-engine-kmp. In 0.60.0, this changed ([ref](https://github.com/charleskorn/kaml/releases/tag/0.60.0)):

> Updated: build with Kotlin 2.0.0
> Updated: use snakeyaml-engine-kmp 3.0.0
> Updated: use kotlinx-serialization 1.7.0

[...]

PR #177, opened 2024-05-11
Refactor StreamReader

aSemy:
Avoid eagerly fetching data.

Use ArrayDeque as a buffer instead of custom array with manual index management.

PR #207, opened 2024-06-28
Use read-only wrapper for codepoints in Mark class instead of making a full array copy

OptimumCode:
This PR should solve the performance degradation introduced with PR #182.

A proper benchmark in the project should prevent such issues in future (I think it would be better to add it in a separate PR).

@krzema12 @aSemy could you please take a look at the PR?

Relates to #204


src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt
```diff
@@ -98,7 +98,7 @@ class StreamReader(
             index = index,
             line = line,
             column = column,
-            codepoints = codePointsWindow.toList(),
+            codepoints = codePointsWindow.toReadOnlyListAdapter(),
```
aSemy, line 101: There's also [`asList()`](https://kotlinlang.org/api/core/kotlin-stdlib/kotlin.collections/as-list.html). I've tried it and I think it works the same. WDYT?

OptimumCode: Yea, indeed. Completely forgot about that

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/exceptions/Mark.kt
```diff
@@ -39,7 +39,7 @@ class Mark @JvmOverloads constructor(
 ) {
 
     @Deprecated("Converted to immutable List<Int>, replace with `codepoints`")
-    val buffer: IntArray = codepoints.toIntArray()
+    val buffer: IntArray by lazy { codepoints.toIntArray() }
```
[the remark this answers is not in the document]
OptimumCode: I agree. Maybe also increase depreciation level to Error?

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt
OptimumCode: Interesting note though:

For the JS target the method `asList` returns a mutable list

```kotlin
public actual inline fun IntArray.asList(): List<Int> {
    return this.unsafeCast<Array<Int>>().asList()
}

public actual fun <T> Array<out T>.asList(): List<T> {
    return ArrayList<T>(this.unsafeCast<Array<Any?>>())
}
```

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/exceptions/Mark.kt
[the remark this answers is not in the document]
aSemy: Thanks for changing it.

Maybe we can change the deprecation level to error later, but for now it's fine as a warning. AFAIK Kaml is the only user of SnakeKMP, so as long as Kaml updates, it's fine.

aSemy (approved):

OptimumCode:
@aSemy Maybe we should remove the performance test since it results in different time execution on different platforms? And there is also #208 PR that should cover that

aSemy:
I'm planning some more refactoring #177, so it'd be useful to have right now, even if it's not very strict. I've relaxed the timing, but yes, proper performance testing will be better.

krzema12:
Thanks for taking care of this @OptimumCode @aSemy! Is it all we need for the patch release?

OptimumCode:
Hi, @krzema12. I believe that should be all we need

aSemy:
Yes, this PR fixes #204.

krzema12:
Ack, I'll prepare the release tentatively on Monday.

PR #209, opened 2024-06-29
Add benchmark project and action to alert performance issues

### LINKED ISSUE #208: Add benchmarks to catch performance degradation on earlier stages
Based on #204 discussion it would be good to have a benchmark running as a part of the tests. It should help to catch any performance degradation before any of the actual users.

I can think of two options:

1. Create a benchmark for JVM that will compare performance for actual SnakeYaml and for this KMP implementation. The results for snakeyaml-engine-kmp MUST NOT be worse than SnakeYaml results with some delta in percentage. This benchmark can be created using [jmh](https://github.com/openjdk/jmh) and executed from tests using `org.openjdk.jmh.runner.Runner`.
2. Create a self-made "benchmark" that will be running for all supported platforms. We need to measure the performance for all platforms on the GitHub runner. After that, we can compare the results for each run with those values and fail tests if values are different for more than some percentage. The [kotlinx-resources](https://github.com/goncalossilva/kotlinx-resources) library can be used to read YAML files required for benchmarks.

There is also a [kotlinx-benchmark](https://github.com/Kotlin/kotlinx-benchmark) library but I don't see how it can be used here because there is no way to execute the generated benchmarks from the test (at least there is no official way to do that - all the generated classes as well as benchmark executors are marked with internal API annotation)

@krzema12 @aSemy What do you think about those options? Do you have any objections or suggestions? Maybe you can other options here

I would think that the second option is better - it provides measurements for all supported platforms and should help to notice any degradation that could be platform-specific.

OptimumCode:
Hi, this is the initial version of the PR. @aSemy could you please take a look?

I faced some issues when adding the benchmark:
1. For JS target I got an error during benchmark generation:
    ```text
    KLIB resolver: Could not find "net.thauvin.erik.urlencoder:urlencoder-lib" in [.../snakeyaml-engine-kmp]
    ```
   I have not figured out the exact reason, but I added this library explicitly into `jsMain` dependencies and it started working. Maybe anyone has seen something like that before...
2. The GitHub action works only with a **single result file**. Because of that, it is a bit tricky to upload results for other platforms (especially because benchmarks for windows and macos must be executed on corresponding runners).
    I think that we could create a reusable workflow that executes all benchmarks available for a particular runner (ubuntu, windows, macos), collects the results (using artifacts or github cache) and uploads all the results in a separate job using matrix build (which contains all the supported targets for that runner). But I don't know how this action will work if those workflows try to commit changes in parallel. I think nothing good will happen... Maybe anyone has other ideas (create a PR to the action repo with multi-file support?).

Here is what the generated page with benchmark results looks like:
https://optimumcode.github.io/snakeyaml-engine-kmp/dev/bench/

To make it work you need to do some [preparations](https://github.com/benchmark-action/github-action-benchmark/blob/master/README.md#charts-on-github-pages-1) first (to allow action publish the result to GitHub pages).

Branch with benchmark results:
https://github.com/OptimumCode/snakeyaml-engine-kmp/tree/gh-pages/dev/bench

Here is an example of PR that introduces a performance issue:
https://github.com/OptimumCode/snakeyaml-engine-kmp/pull/1

Also, every workflow run generates a summary where you can see how the benchmark results have changed:
https://github.com/OptimumCode/snakeyaml-engine-kmp/actions/runs/9725885272?pr=1

Resolves #208


OptimumCode:
And I think this PR should be merged after #207 (I will rebase branch once it is merged)

OptimumCode:
There is a [suggestion](https://github.com/benchmark-action/github-action-benchmark/issues/69#issuecomment-1069394041) on how multiple benchmark results can be combined and processed together.

OptimumCode:
I have managed to join results from multiple runs in a single file and uploaded them using benchmark-action. This is [workflow step](https://github.com/OptimumCode/json-schema-validator/blob/13ab8df1c1e0c2d410b0ce115db9db9f22fd470d/.github/workflows/benchmark.yml#L158) to do that (and the [result](https://optimumcode.github.io/json-schema-validator/dev/bench/)). Could the same approach be used here? What do you think?

aSemy:
Thanks for this! I will take a look.

benchmark/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/benchmark/LoadingTimeBenchmark.kt
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    @Benchmark
+    fun loadsOpenAiSchema(): Map<*, *> {
+        return with(FILE_SYSTEM) {
+            openReadOnly(openAiYamlPath.toPath(normalize = true)).use { handle ->
+                handle.source().buffer().use { source ->
+                    // TODO: there is a Load class in JVM sources that can handle all of it
+                    //       but it is not available for common code.
+                    //       Probably, it should be moved from JVM sources to common sources.
```
aSemy, line 39: Good idea, I'll try to commonize Load.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,51 @@
...
+    - name: Find JVM benchmark result
+      id: results
+      run: echo "report=$(find benchmark/build/reports/benchmarks/main/ -name jvm.json -type f)" >> $GITHUB_OUTPUT
+    - name: Store benchmark result
+      uses: benchmark-action/github-action-benchmark@v1
+      with:
+        name: SnakeKMP benchmark
+        tool: 'jmh'
+        output-file-path: ${{ steps.results.outputs.report }}
+        alert-comment-cc-users: "@krzema12,@aSemy"
```
aSemy, line 45:
```suggestion
```

The action leaving a comment in the PR is fine, I don't need to be pinged, and I assume `@krzema12` doesn't need to be either.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,51 @@
+name: Run benchmark for parser
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+
+permissions:
+  # deployments permission to deploy GitHub pages website
+  deployments: write
```
aSemy, line 10: Just a thought: maybe writing to GitHub pages should only be enabled when running on the main branch? If I understand the benchmark action, then the benchmarks should only be updated when running on main, but I wonder if an additional guard would help.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,51 @@
...
+  # contents permission to update benchmark contents in gh-pages branch
+  contents: write
+  # pull-requests permission to create comments on PR in case of alert
+  pull-requests: write
```
aSemy, line 14: Thank you for adding explanations for the permissions 🙏

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,51 @@
+name: Run benchmark for parser
```
aSemy, line 1:
```suggestion
name: Run Benchmarks
```

nitpick

benchmark/build.gradle.kts
```diff
@@ -0,0 +1,98 @@
...
+                implementation("com.squareup.okio:okio")
+            }
+        }
+
+        jsMain {
+            dependencies {
+                implementation("net.thauvin.erik.urlencoder:urlencoder-lib:1.5.0") {
+                    because("error during benchmark generation: e:\n" +
+                        "KLIB resolver: Could not find \"net.thauvin.erik.urlencoder:urlencoder-lib\" in [.../snakeyaml-engine-kmp]")
+                }
```
aSemy, line 64: That's an unexpected error! I wonder why it doesn't happen on other targets. I'll take a look... 

Maybe snake-kmp needs to mark it as `api()`?

settings.gradle.kts
```diff
@@ -1,5 +1,7 @@
 rootProject.name = "snakeyaml-engine-kmp"
 
+include("benchmark")
```
aSemy, line 3: I'm a big fan of using a common prefix in subprojects, so that they are visually grouped together in the file list. What about renaming the benchmarks to `snake-kmp-benchmarks`?

benchmark/build.gradle.kts
OptimumCode: I think this might be related to how the benchmark plugin resolves dependencies when compiling generated benchmarks. But I believe snake-kmp should not mark this dependency as `api` because it does not expose this library in public API (even if it fixes this problem for some reason)

settings.gradle.kts
OptimumCode: I have no preferences here) Will rename the project

benchmark/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/benchmark/LoadingTimeBenchmark.kt
aSemy: https://github.com/krzema12/snakeyaml-engine-kmp/pull/210

OptimumCode:
Hi, @aSemy. I want to rebase the branch on the latest `main` state. Please, let me know if you are doing some changes - I will do a merge instead of rebase in this case (but I would prefer a liner history)

aSemy:
I'm not working on this branc, go ahead and rebase if you'd like. We'll squash merge the branch into main, so I wouldn't worry too much about rebasing vs merging.

.github/workflows/benchmark.yml
OptimumCode: But still, it would be possible to modify the PR workflow and push results from PR (it just will be more obvious because the permissions will be modified)

benchmark/build.gradle.kts
OptimumCode: You are actually right... if you declare the dependency to that library as `api`, the benchmark for JS compiles without any issues. But it is not right to include a library as `api` if it is not used in public API. I will raise a question in kotlinx-benchmark repo - maybe they can help with that

.github/workflows/benchmark.yml
OptimumCode: There is already an issue for that: https://github.com/Kotlin/kotlinx-benchmark/issues/185
`Re-declaration of dependencies.`

benchmark/build.gradle.kts
OptimumCode: There is already an issue for that: https://github.com/Kotlin/kotlinx-benchmark/issues/185
`Re-declaration of dependencies.`

OptimumCode:
Hi, @aSemy
I merged the results for all platforms and they are all available on one page (https://optimumcode.github.io/snakeyaml-engine-kmp/dev/bench/). Does it look okay? Could you please check when you have time?

OptimumCode:
Hi, @krzema12. Could you please also take a look at the PR?

The example can be found in the forked repo: https://github.com/OptimumCode/snakeyaml-engine-kmp/pull/1
https://optimumcode.github.io/snakeyaml-engine-kmp/dev/bench/

krzema12:
@OptimumCode thanks for this, it looks good in general, but we need to sort out one problem. From what I see, the new workflow publishes the report via GitHub Pages, but we already use it to publish API reference: https://github.com/krzema12/snakeyaml-engine-kmp/blob/main/.github/workflows/publish-site.yml.

One idea is to deploy both the docs and the performance report using GH Pages, but under different paths. Then we'd also need to ensure that both workflows don't override each other's results.

I'm gonna sleep on this problem, maybe a better solution will appear, but I'm also open to your ideas!

OptimumCode:
Hi, @krzema12. Thank you for pointing out that you already use the github pages in another workflow. I looked into documentation for **actions/upload-pages-artifact**, **actions/deploy-pages** and **benchmark-action/github-action-benchmark** actions - seems like they won't work together without some manual actions...
actions/deploy-pages expects a single artifact with **all content** for github-pages that fully overrides the previously deployed data. This would mean that results from benchmark-action/github-action-benchmark **will override** the previously deployed documentation.

I have one thought here: benchmark-action/github-action-benchmark action supports [publishing results into another repository](https://github.com/benchmark-action/github-action-benchmark?tab=readme-ov-file#gh-repository). What do you think about creating a **snakeyaml-engine-kmp-benchmarks** (or any other name you prefer) repository and publishing results into this repo?

krzema12:
I'm fine with the separate repo! I'll let you know once created. Do I need to set up any extra perms anywhere?

OptimumCode:
You will need to create a personal access token with **deployment** and **contents** write permissions for this new repo and add it to the secrets here

krzema12:
@OptimumCode done! The secret is called `PUBLISH_BENCHMARK_RESULTS` and will expire in a year (it's impossible to set no expiration date). The dedicated repo is here: https://github.com/krzema12/snakeyaml-engine-kmp-benchmarks.

OptimumCode:
Thank you, @krzema12. I will make necessary changes and let you know once it is done

OptimumCode:
Hi, @krzema12. Could you please advise which branch have you configured for GitHub pages in this new repo? If you haven't could you please create a branch (gh-pages is a default value expected by the action) and configure GitHub pages to serve from this branch?
I completely forgot about this when you asked for what is needed.

krzema12:
@OptimumCode I've just configured gh-pages branch.

OptimumCode:
Hi, @krzema12. Could you please check if the name of the secret is correct? Either the name is not correct or the secret is not visible in pr workflow...

One more thing, could you please add pull_request write permission to this token for main repo (not for repo with benchmark results)? This is needed to generate an alert comment

krzema12:
Hi @OptimumCode!

> Could you please check if the name of the secret is correct? Either the name is not correct or the secret is not visible in pr workflow...

Yeah, the name is correct, checked by copy-pasting between GitHub's UI and workflow's definition:

Screenshot 2024-07-26 at 09 03 58

What I can do is regenerate the token and update the secret, to make sure I correctly pasted it. I've just done it - let's try now! 

> One more thing, could you please add pull_request write permission to this token for main repo (not for repo with benchmark results)? This is needed to generate an alert comment

I think you mean ticking the "Allow GitHub Actions to create and approve pull requests" checkbox, is it correct? If yes, I've just done it!

OptimumCode:
> I think you mean ticking the "Allow GitHub Actions to create and approve pull requests" checkbox, is it correct? If yes, I've just done it!

I am not sure this will do the trick. I was talking about token permissions:

image

krzema12:
@OptimumCode what you show on the screenshot looks like configuring a personal access token's perms - I see this when configuring the PAT for publishing the report to the other repo.

For the GitHub token, I'm only aware of the setting I mentioned, and setting fine-grained perms via YAML, see https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/

I'll try digging deeper into why things don't work as expected here.

OptimumCode:
Yes, it is PAT token configuration. We use it instead of GitHub token in benchmark action and in order to generate alert comments this token should have permissions for that (I will remove permissions confirmation from workflow itself as we use the PAT token)

> I'll try digging deeper into why things don't work as expected here.

Please let me know if I can help here anyhow (right now I have a limited access to the internet but in few hours I will be fully available)

krzema12:
@OptimumCode ok, got it now! I somehow thought that the GH token is **also** used, but now I see that we switched to PAT to be able to write to another repo. I've adjusted the PAT's perms, let's see now.

.github/workflows/benchmark.yml
```diff
@@ -0,0 +1,101 @@
...
+        ls ${{ env.RESULTS_DIR }}/*.json
+        jq -s '[ .[] | .[] ]' ${{ env.RESULTS_DIR }}/*.json > $AGGREGATED_REPORT
+        echo "report=$AGGREGATED_REPORT" >> $GITHUB_OUTPUT
+    - name: Store benchmark result
+      uses: benchmark-action/github-action-benchmark@v1
+      with:
+        name: SnakeKMP benchmarks
+        tool: 'jmh'
+        output-file-path: ${{ steps.prep.outputs.report }}
+        comment-on-alert: true
```
krzema12, line 94: @OptimumCode if the token issue is still blocking us, how about setting `comment-on-alert` to false for now, and relying on the fact that the job should fail?

OptimumCode: I have no objections to that. But I think if the secret with PAT is not visible in the workflow we won't be able to clone repo with results and push new results back. And this is the main issue here at the moment

krzema12: True! This problem seems to be with GitHub Actions, not this particular action. I'll play with this, e. g. by creating different kinds of secrets, although I think the action secret should work here.

OptimumCode: Me too. Probably it is happening because I am not a contributor - probably GitHub doesn't expose secrets in such cases. Or maybe because this is a PR from a fork. I will also dig into this

krzema12: > Or maybe because this is a PR from a fork

Oh, right! Let me make you a contributor, then you can move your branch to this repo and perhaps recreate the PR. Let's try it!

OptimumCode: Thank you! Once I have access to the computer I will move the branch.

RE secrets: this is what I have found

> With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

https://docs.github.com/en/actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow

Probably, it can be solved with an environment but I am not sure

OptimumCode: Looks like this might be a solution: https://github.com/community/community/discussions/50161#discussioncomment-5824158.

If we don't configure the execution that works for PRs from forks there is not much point in having this job, I think... I will do required changes in the workflow. @krzema12 could you please create an environment that I could use in the workflow? Also could you please add this PAT to the environment you create?

OptimumCode: The environment is just a safety measure

OptimumCode: Actually, on a second thought, we might not need an environment... I think we can split the workflow into two separate workflows:
- One runs benchmark and stores results as artifacts
- another downloads results and executes benchmark action (which requires access to secrets). This can be done using **workflow_run** trigger

As a result, code from pr is executed in safe environment without access to secrets and results processing is executed with access to secrets

OptimumCode: @krzema12 what do you think about this idea? It sounds complicated but as a result, provides a bit more safety

krzema12: How about we go with the current approach, as it already gives us some value when making changes to the library (right now it's mostly @aSemy who e. g. does some refactoring), and if the volume of external contributions rise, we'll consider splitting the workflow like you propose?

OptimumCode: This is completely up to you) In this case, I will move the branch and recreate the PR

OptimumCode:
Superseded by #215

PR #210, opened 2024-07-01
Commonize Load

aSemy:
- convert Load from Kotlin JVM to Common
- Create LoadCommon for shared Load functions

⚠️ Do not squash merge. `Load.kt` was renamed to `Load.jvm.kt`. In order to preserve Git history, this PR should not be squashed.


src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/api/Load.kt
```diff
@@ -0,0 +1,61 @@
...
+    fun loadOne(string: String): Any?
+
+    /**
+     * Parse a YAML document and create an instance of an object.
+     *
+     * @param source YAML data to load
+     * @return parsed instance
+     * @throws it.krzeminski.snakeyaml.engine.kmp.exceptions.YamlEngineException if the YAML is not valid
+     */
+    fun loadOne(source: Source): Any?
```
OptimumCode, line 39: suggestion: I think the `okio` dependency should be added to `commonMain` dependencies as `api` instead of `implementation` since it is exposed in public API

aSemy: Ah yes, good point. I think it'd be best not to expose Okio as a required dependency. Ideally the official library https://github.com/Kotlin/kotlinx-io/ will become more fully-featured, and we can migrate from Okio. So I'll make these functions internal.

src/jvmMain/java/it/krzeminski/snakeyaml/engine/kmp/api/Load.jvm.kt
krzema12, line 1: Before I review further, let's discuss adding API dump: https://github.com/krzema12/snakeyaml-engine-kmp/pull/213.

krzema12 (approved):

PR #213, opened 2024-07-22
Add JVM API dump

krzema12:
It will let us get more confidence that no API changes (wrt. signatures) are introduced whenever we e.g. refactor stuff, and help us review API changes if we make them.


aSemy (approved):

aSemy:
It could be a good idea to add a check (in another PR) to make sure we don't expose any Okio types in the public ABI? Just a simple `"Okio" !in file("snake-kmp.api").readText()` check.

PR #214, opened 2024-07-22
Ensure no Okio in API

krzema12:


krzema12:
So, Okio is actually present in the API dump. Next step: review the occurrences and check if the definitions can be made internal/private.

aSemy:
I'd prefer it if this check could be done locally. It'd be convenient to be able to run all tests locally, so I could be more confident that the tests would pass on CI.

krzema12:
Ah, right! Having it as a unit test makes sense, not sure why I implemented it this way.

krzema12:
@aSemy just wondering: if we switch to kotlinx.io, would we actually let it be present in the API? It'd make sense to me.

aSemy:
> @aSemy just wondering: if we switch to kotlinx.io, would we actually let it be present in the API? It'd make sense to me.

Yes, I think so, assuming that kotlinx-io becomes the defacto standard.

TBH I think right now that Okio is the defacto standard, and we should consider just requiring it as a dependency, and using it in the public ABI. If/when kotlinx-io is released as stable, we could migrate to it.

krzema12:
Ok, so I think we should discard this PR.

PR #216, opened 2024-07-26
README: mention performance benchmarks

krzema12:


OptimumCode (approved):
Looks good to me)

btw we can add another benchmark to compare results with the original SnakeYaml library

krzema12:
Good idea! I'll try to add it so that it's on the same plot as the KMP version.

PR #219, opened 2024-07-28
Configure separate repo for benchmark testing

krzema12:
The goal is to be able to test changes in the performance report generation within PR. The report will be pushed to a separate repo, see https://krzema12.github.io/snakeyaml-engine-kmp-benchmarks-testing/dev/bench/

### DIFF
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -88,7 +88,7 @@ jobs:
         summary-always: true
         alert-threshold: '150%'
         fail-threshold: '200%'
-        gh-repository: ${{ github.event_name == 'pull_request' && 'github.com/krzema12/snakeyaml-engine-kmp-benchmarks' || 'github.com/krzema12/snakeyaml-engine-kmp-benchmarks-testing' }}
+        gh-repository: ${{ github.event_name != 'pull_request' && 'github.com/krzema12/snakeyaml-engine-kmp-benchmarks' || 'github.com/krzema12/snakeyaml-engine-kmp-benchmarks-testing' }}
         github-token: ${{ secrets.PUBLISH_BENCHMARK_RESULTS }}
         # Push and deploy GitHub pages branch automatically only if run in main repo
         auto-push: ${{ github.repository == 'krzema12/snakeyaml-engine-kmp'}}
--- a/.github/workflows/benchmark.yml
+++ b/.github/workflows/benchmark.yml
@@ -88,7 +88,7 @@ jobs:
         summary-always: true
         alert-threshold: '150%'
         fail-threshold: '200%'
-        gh-repository: github.com/krzema12/snakeyaml-engine-kmp-benchmarks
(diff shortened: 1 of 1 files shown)


.github/workflows/benchmark.yml
```diff
@@ -88,7 +88,7 @@ jobs:
         summary-always: true
         alert-threshold: '150%'
         fail-threshold: '200%'
-        gh-repository: github.com/krzema12/snakeyaml-engine-kmp-benchmarks
+        gh-repository: ${{ github.event_name != 'pull_request' && 'github.com/krzema12/snakeyaml-engine-kmp-benchmarks' || 'github.com/krzema12/snakeyaml-engine-kmp-benchmarks-testing' }}
         github-token: ${{ secrets.PUBLISH_BENCHMARK_RESULTS }}
-        # Push and deploy GitHub pages branch automatically only if run in main repo and not in PR
-        auto-push: ${{ github.repository == 'krzema12/snakeyaml-engine-kmp' && github.event_name != 'pull_request' }}
+        # Push and deploy GitHub pages branch automatically only if run in main repo
+        auto-push: ${{ github.repository == 'krzema12/snakeyaml-engine-kmp'}}
```
OptimumCode, line 94: note: It looks a bit strange to me - on PR we publish results to the test repo and **compare results with data from the test repo** and when PR is merged the results are published into the main repo. I think it would be better to keep only the main repo in the benchmark action.

But we can add a new job, that will duplicate the results from the main benchmark repository into test repository. As a result, the benchmark action will always compare/publish results with/into the main repository. And the test repository will be a copy of the main repository. And you will be able to experiment with HTML in test repository using the same data that is stored in the main repo

krzema12: > and compare results with data from the test repo

Great catch! This innocent change could actually render part of our new mechanism useless.

Regarding copying the main repo to the test repo, I'm wondering when it would have to happen. Before running the benchmark action, both for pushing to `main` and in PRs? Then the only problematic case would be two or more PRs being pushed to at roughly the same time, the they'd overwrite their reports in the test repo. But I think it's bearable.

OptimumCode: **question:** Do we really want to store results from each PR to the test repo?

I believe the main point of having this test repo is to be able to modify the index.html file to change how those charts are displayed. Please, correct me if I am wrong.
If I am right, I think we can copy results only when PR is merged and they are published into the main repo. In this case, we should not have a problem with parallel commits.

**suggestion:** What do you think about an alternative approach?
We could create a workflow in the test repository that is triggered by cron and does the following:
- clones the main repo with benchmark results
- checks if the data is different from the one is already stored in the test repo
- updates data if the data from the main repo is different from the data in test repo

In this case, we:
- don't add any information about the test repo here - this is just a test repo and there is no need to link it to this repository
- don't have a problem with parallel commits - the workflow is scheduled by cron and only one workflow is executed at a time

krzema12: Thanks for the ideas. Too much complexity, I'm discarding this PR 😊

PR #221, opened 2024-07-31
Use Kotlin DSL for benchmark workflow

krzema12:
Use [github-workflows-kt](https://github.com/typesafegithub/github-workflows-kt) to implement the workflow in Kotlin, increasing maintainability and readability.

Note: proper, type-safe matrix strategy support is yet to be implemented, it's tracked in https://github.com/typesafegithub/github-workflows-kt/issues/368.

# Testing done
1. Compared what's logged in the "Store benchmark result" step as "Data", and there are no differences.
  * version from `main` branch: https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/10187305280/job/28181268567
  * version from this branch: https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/10244555647/job/28338076917?pr=221
2. Compared with https://www.yamldiff.com/. Most important differences:
  * the new version has a consistency check which is required to ensure that the YAML reflects what's described in Kotlin
  * the new version doesn't have the reports preprocessing step implemented directly, and instead delegates it to the Kotlin script itself
  * instead of using step outputs or env vars (so effectively to not repeat ourselves with the file paths), I just used Kotlin's constants
  * github-workflows-kt proactively adds step IDs to be able to refer to them whenever needed
   SemanticDiff

### DIFF
--- a/.github/workflows/benchmark.main.kts
+++ b/.github/workflows/benchmark.main.kts
@@ -1,0 +1,161 @@
+#!/usr/bin/env kotlin
+@file:Repository("https://repo1.maven.org/maven2/")
+@file:DependsOn("io.github.typesafegithub:github-workflows-kt:2.3.0")
+@file:DependsOn("org.jetbrains.kotlinx:kotlinx-serialization-json:1.7.1")
+
+@file:Repository("https://bindings.krzeminski.it/")
+@file:DependsOn("actions:checkout:v4")
+@file:DependsOn("actions:download-artifact:v4")
+@file:DependsOn("actions:upload-artifact:v4")
+@file:DependsOn("actions:setup-java:v4")
+@file:DependsOn("gradle:actions__wrapper-validation:v4")
+@file:DependsOn("gradle:actions__setup-gradle:v4")
+@file:DependsOn("benchmark-action:github-action-benchmark:v1")
+
+import io.github.typesafegithub.workflows.actions.actions.Checkout
+import io.github.typesafegithub.workflows.actions.actions.DownloadArtifact
+import io.github.typesafegithub.workflows.actions.actions.UploadArtifact
+import io.github.typesafegithub.workflows.actions.actions.SetupJava
+import io.github.typesafegithub.workflows.actions.benchmarkaction.GithubActionBenchmark
+import io.github.typesafegithub.workflows.actions.gradle.ActionsSetupGradle
+import io.github.typesafegithub.workflows.actions.gradle.ActionsWrapperValidation
+import io.github.typesafegithub.workflows.annotations.ExperimentalKotlinLogicStep
+import io.github.typesafegithub.workflows.domain.Concurrency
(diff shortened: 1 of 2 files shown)


krzema12:
Blocked on https://github.com/typesafegithub/github-workflows-kt/issues/1562

.github/workflows/benchmark.main.kts
```diff
@@ -0,0 +1,161 @@
...
+            name = "Prepare and join benchmark reports",
+        ) {
+            val mergedReports = File(RESULTS_DIR).walk()
+                .filter { it.extension == "json" }
+                .flatMap {
+                    val reportForPlatform = it.readText()
+                        // Trim lengthy Kotlin package and data file path to make benchmark name more readable
+                        .replace("it.krzeminski.snakeyaml.engine.kmp.benchmark", it.nameWithoutExtension)
+                        .replace(Regex("\"[^\"]+data"), "\"data")
+                    Json.parseToJsonElement(reportForPlatform).jsonArray
```
OptimumCode, line 136: note: In my opinion, it would be better (and safer) to transform the `JsonElement` rather then replacing values in the file with regex. But if the result is the same, I have no objections to this approach.

.github/workflows/benchmark.yaml
```diff
@@ -1,94 +1,103 @@
...
+        summary-always: 'true'
+        comment-on-alert: 'true'
         alert-threshold: '150%'
         fail-threshold: '200%'
-        gh-repository: github.com/krzema12/snakeyaml-engine-kmp-benchmarks
-        github-token: ${{ secrets.PUBLISH_BENCHMARK_RESULTS }}
-        # Push and deploy GitHub pages branch automatically only if run in main repo and not in PR
-        auto-push: ${{ github.repository == 'krzema12/snakeyaml-engine-kmp' && github.event_name != 'pull_request' }}
+        github-token: '${{ secrets.PUBLISH_BENCHMARK_RESULTS }}'
+        auto-push: '${{ github.repository == ''krzema12/snakeyaml-engine-kmp'' && github.event_name != ''pull_request'' }}'
```
OptimumCode, line 103: question: Is it okay that the line has two `'` quotes around string values?

OptimumCode (approved):

.github/workflows/benchmark.main.kts
krzema12: You're right, I thought of it, but then realized that the values to be replaced sit three levels down in the JSON hierarchy, so it would make this code pretty complicated. That's why I went with this simpler approach, even though it may be a bit unsafe. Happy to revisit later if it turns out to be problematic.

.github/workflows/benchmark.yaml
krzema12: Yes, it's a way to escape single quotes in a string if we use single quotes for the sting in YAML. It's handled entirely by the library, and "just works" 😊

PR #228, opened 2024-08-20
Add GitHub workflow to publish a release

### LINKED ISSUE #94: Create a release GitHub workflow
E.g. after pushing a tag, a release process of building for all targets (especially Mac OS) should start.
Currently I have releasing configured on my legacy Ubuntu machine, which effectively makes this library not provide OS X targets.

OptimumCode:
Part of #94

The workflow uses [nexus-actions](https://github.com/nexus-actions) to manage the staging repository. The publication was updated to use the new URL for deployment when a staging repository ID is available.

The following secrets should be added to the project:

- SONATYPE_USERNAME
- SONATYPE_PASSWORD
- SONATYPE_STAGING_PROFILE_ID (can be obtained using this link: https://oss.sonatype.org/#stagingProfiles;$staginProfileId)
- SIGNING_KEY_ID
- SIGNING_KEY
- SIGNING_PASSWORD

@aSemy @krzema12 Could you please take a look at the changes when you have time? Unfortunately, I don't have an OSS account (only OSS1) and cannot test the workflow by myself. But if you have any issues when trying to publish a new release I will be happy to help

### DIFF
--- a/.github/workflows/release.main.kts
+++ b/.github/workflows/release.main.kts
@@ -40,56 +40,58 @@ workflow(
     ),
     sourceFile = __FILE__,
 ) {
-    val stagingRepoJob = job(
-        id = "create-staging-repo",
-        name = "Create staging repository",
-        runsOn = RunnerType.UbuntuLatest,
-        outputs = object : JobOutputs() {
-            var repositoryId: String by output()
-        }
-    ) {
-        val createRepo = uses(
-            action = CreateNexusStagingRepo(
-                username = expr { SONATYPE_USERNAME },
-                password = expr { SONATYPE_PASSWORD },
-                stagingProfileId = expr { SONATYPE_STAGING_PROFILE_ID },
+    val stagingRepoJob =
+        job(
+            id = "create-staging-repo",
+            name = "Create staging repository",
+            runsOn = RunnerType.UbuntuLatest,
+            outputs = object : JobOutputs() {
+                var repositoryId: String by output()
+            }
+        ) {
+            val createRepo = uses(
+                action = CreateNexusStagingRepo(
+                    username = expr { SONATYPE_USERNAME },
+                    password = expr { SONATYPE_PASSWORD },
+                    stagingProfileId = expr { SONATYPE_STAGING_PROFILE_ID },
+                )
             )
-        )
-        jobOutputs.repositoryId = createRepo.outputs.repositoryId
-    }
-    val publishJob = job(
-        id = "publish-artifacts",
-        runsOn = RunnerType.MacOSLatest,
-        needs = listOf(stagingRepoJob),
-    ) {
-        uses(action = Checkout())
-        uses(
-            name = "Set up JDK",
-            action = SetupJava(
-                javaVersion = "11",
-                distribution = SetupJava.Distribution.Zulu,
-                cache = SetupJava.BuildPlatform.Gradle,
-            ),
-        )
-        uses(
-            name = "Cache Kotlin Konan",
-            action = Cache(
-                path = listOf(
-                    "~/.konan/**/*",
+            jobOutputs.repositoryId = createRepo.outputs.repositoryId
+        }
+    val publishJob =
+        job(
+            id = "publish-artifacts",
+            runsOn = RunnerType.MacOSLatest,
+            needs = listOf(stagingRepoJob),
+        ) {
+            uses(action = Checkout())
+            uses(
+                name = "Set up JDK",
+                action = SetupJava(
+                    javaVersion = "11",
+                    distribution = SetupJava.Distribution.Zulu,
+                    cache = SetupJava.BuildPlatform.Gradle,
                 ),
-                key = "kotlin-konan-${expr { runner.os }}",
-            ),
-        )
-        uses(
-            name = "Set up Gradle",
-            action = ActionsSetupGradle(
-                gradleVersion = "wrapper",
-            ),
-        )
-        run(
-            name = "Publish",
-            command = "./gradlew publishAllPublicationsToSonatypeReleaseRepository",
-            env =
+            )
+            uses(
+                name = "Cache Kotlin Konan",
+                action = Cache(
+                    path = listOf(
+                        "~/.konan/**/*",
+                    ),
+                    key = "kotlin-konan-${expr { runner.os }}",
+                ),
+            )
+            uses(
+                name = "Set up Gradle",
+                action = ActionsSetupGradle(
+                    gradleVersion = "wrapper",
+                ),
+            )
+            run(
+                name = "Publish",
+                command = "./gradlew publishAllPublicationsToSonatypeReleaseRepository",
+                env =
                 mapOf(
                     "ORG_GRADLE_PROJECT_snake-kmp.ossrhUsername" to expr { SONATYPE_USERNAME },
                     "ORG_GRADLE_PROJECT_snake-kmp.ossrhPassword" to expr { SONATYPE_PASSWORD },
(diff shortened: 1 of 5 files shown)


krzema12:
Ack, thank you 🙇 In my queue.

.github/workflows/release.main.kts
```diff
@@ -0,0 +1,134 @@
...
+            )
+            uses(
+                name = "Set up Gradle",
+                action = ActionsSetupGradle(
+                    gradleVersion = "wrapper",
+                ),
+            )
+            run(
+                name = "Publish",
+                command = "./gradlew publishAllPublicationsToSonatypeReleaseRepository",
```
aSemy, line 93:
```suggestion
                command = "./gradlew publishAllPublicationsToSonatypeReleaseRepository --stacktrace",
```

Suggestion: Log stacktrace, because it can really help with determining problems if publishing fails.

.github/workflows/release.main.kts
```diff
@@ -0,0 +1,134 @@
...
+                    ),
+                    key = "kotlin-konan-${expr { runner.os }}",
+                ),
+            )
+            uses(
+                name = "Set up Gradle",
+                action = ActionsSetupGradle(
+                    gradleVersion = "wrapper",
+                ),
+            )
```
OptimumCode, line 90: issue: Just realized that the caching won't work because the `build` workflow uses a different Gradle action. Do you mind changing it in the build workflow to the same one as here?

.github/workflows/release.main.kts
```diff
@@ -0,0 +1,134 @@
...
+                    "ORG_GRADLE_PROJECT_snake-kmp.signing.key" to expr { SIGNING_KEY },
+                    "ORG_GRADLE_PROJECT_snake-kmp.signing.password" to expr { SIGNING_PASSWORD },
+                )
+            )
+        }
+
+    job(
+        id = "close-staging-repo",
+        runsOn = RunnerType.UbuntuLatest,
+        condition = expr { publishJob.result.eq(AbstractResult.Status.Success) },
```
aSemy, line 109:
```suggestion
        condition = expr { publishJob.result.eq(AbstractResult.Status.Success) },
```

Question for @krzema12: It's a little awkward reading `.eq` (and `.neq` below). Could `==` and `!=` be used? Or does something interfere with it?

aSemy (approved):
Incredible, thanks!

.github/workflows/release.main.kts
aSemy: Ah, good catch. Yes, I think `workflows
/build.main.kts` should be updated to be more like this workflow, which (I think) correctly follows the newer Gradle Actions set up.

.github/workflows/release.main.kts
OptimumCode: This is not a question to me, but I will try to share what I know: `==` and `!=` delegates to `equals` call. The signature of `equals` method is `fun equals(other: Any?): Boolean`. So, we cannot return a `String` from the method. Therefore, we cannot use it in `expr` function

.github/workflows/release.main.kts
```diff
@@ -0,0 +1,134 @@
...
+            )
+            uses(
+                name = "Set up Gradle",
+                action = ActionsSetupGradle(
+                    gradleVersion = "wrapper",
+                ),
+            )
+            run(
+                name = "Publish",
+                command = "./gradlew publishAllPublicationsToSonatypeReleaseRepository --stacktrace",
```
OptimumCode, line 93: question: What do you think about adding the `--info` option here as well? It should help to investigate what was published and where (in case of issues with the publication)

aSemy: `--info` tends to be a little too noisy for everyday use.

When it comes to publishing, usually the Gradle tasks are stable and run correctly. It's usually configuring the correct properties in build scripts and Sonatype that are troublesome, and _hopefully_ `--stacktrace` provides enough help for that.

.github/workflows/release.main.kts
krzema12: Like @OptimumCode wrote: we want to build a string here. GitHub Actions are very "stringy". I'm working on being able to specify such conditions in Kotlin, but we're not there yet, and it will for sure be more costly performance-wise.

.github/workflows/release.main.kts
```diff
@@ -0,0 +1,134 @@
...
+@file:DependsOn("nexus-actions:create-nexus-staging-repo:v1")
+@file:DependsOn("nexus-actions:release-nexus-staging-repo:v1")
+@file:DependsOn("nexus-actions:drop-nexus-staging-repo:v1")
+
+import io.github.typesafegithub.workflows.actions.actions.Cache
+import io.github.typesafegithub.workflows.actions.actions.Checkout
+import io.github.typesafegithub.workflows.actions.actions.SetupJava
+import io.github.typesafegithub.workflows.actions.gradle.ActionsSetupGradle
+import io.github.typesafegithub.workflows.actions.nexusactions.CreateNexusStagingRepo
+import io.github.typesafegithub.workflows.actions.nexusactions.DropNexusStagingRepoV1
```
krzema12, line 19: This is a bundled action binding, and it will be gone in the next version of the lib (in 2 weeks) 😊 Could we switch to the binding provided by the Maven-compatible server? The action provides the typings so in fact the only required changes are:

```suggestion
import io.github.typesafegithub.workflows.actions.nexusactions.DropNexusStagingRepo
```

and using this class in the workflow.

.github/workflows/release.main.kts
```diff
@@ -0,0 +1,134 @@
...
+import io.github.typesafegithub.workflows.actions.nexusactions.ReleaseNexusStagingRepo
+import io.github.typesafegithub.workflows.domain.AbstractResult
+import io.github.typesafegithub.workflows.domain.JobOutputs
+import io.github.typesafegithub.workflows.domain.RunnerType
+import io.github.typesafegithub.workflows.domain.triggers.WorkflowDispatch
+import io.github.typesafegithub.workflows.dsl.expressions.Contexts
+import io.github.typesafegithub.workflows.dsl.expressions.expr
+import io.github.typesafegithub.workflows.dsl.workflow
+
+val SONATYPE_USERNAME by Contexts.secrets
```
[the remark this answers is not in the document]
krzema12: User names are now pseudorandom tokens: https://central.sonatype.org/publish/generate-token/

krzema12:
Great job, thanks :) I'll configure all the secrets within a day or two, and give it a spin to publish some dummy patch version.

krzema12:
I see that the benchmarks workflow is failing, doesn't seem related to this PR, so I'll take a look as well.

OptimumCode:
> I see that the benchmarks workflow is failing, doesn't seem related to this PR, so I'll take a look as well.

It cannot import the action to validate the Gradle wrapper. It only finds the class with `_Untyped` suffix.

krzema12:
> > I see that the benchmarks workflow is failing, doesn't seem related to this PR, so I'll take a look as well.
> 
> It cannot import the action to validate the Gradle wrapper. It only finds the class with `_Untyped` suffix.

Ah ok, so this breaking change is by design: https://kotlinlang.slack.com/archives/C02UUATR7RC/p1723929533809109
I'll take care of it.

.github/workflows/release.main.kts
OptimumCode: Sure, will do

krzema12 (approved):

krzema12:
Thanks @OptimumCode, merging so that the workflow appears in the list of workflows, and I can test it with the secrets set.

OptimumCode:
Thank you @krzema12. Let me know if you have any issues with the publication

PR #243, opened 2024-09-23
Remove irrelevant JavaDoc

Vampire:
Strings are never UTF-8 encoded.
Strings are a sequence of Unicode characters.
If these Unicode characters are encoded into a sequence of bytes
or decoded from a sequence of bytes, then encodings like UTF-8 come into play.

Same for BOM, they are just not relevant when we talk about Strings or Readers,
they are only relevant if starting a stream of bytes, otherwise they are just content.


krzema12:
It makes perfect sense to me when you explain it this way, but on the other hand I recall we had some encoding-related issues/changes - @aSemy can you weigh in if the adjustments make sense to you as well (just docs, but good to have common understanding here)?

PR #255, opened 2024-10-16
Add Java 8 support

rbraeunlich:
Closes #202

PR #259, opened 2024-10-26
Switch JDK distribution to Temurin

krzema12:
See https://github.com/krzema12/snakeyaml-engine-kmp/pull/255#discussion_r1817840117

Apparently Temurin is one of the most recommended ones.


krzema12:
Self-approved since we basically agreed on this in the other PR.

Issue #260, opened 2024-10-28
Track changes to be synced from snakeyaml-engine


krzema12:
This project was forked from https://bitbucket.org/snakeyaml/snakeyaml-engine/. Even though it was ported to Kotlin, the general structure is still very close to snakeyaml-engine. Right now we sync changes from that project periodically, but there's no set cadence nor mechanisms that help us.

As the first step, for the sake of monitoring if there are any changes to be ported over to our project, I suggest the following approach:
1. Have a file with commit hash of snakeyaml-engine pointing to the latest commit that was already ported to snakeyaml-engine-kmp.
2. Have a workflow that periodically (every day/week?) calculates how many commits there are between the tip of the master branch and the commit stored in the file in our project.
3. Surface this number in e.g. a badge in README, with a note that ideally it should be 0, and with notes how to proceed if it's not 0.

PR #262, opened 2024-10-28
Add self-updating badge with commits to snakeyaml-engine

krzema12:
Part of #260.

Thanks to this, we'll be able to easily see if there are some changes to sync from the upstream.
(Don't look at the messy git history, it's irrelevant.)


.github/workflows/check-upstream.main.kts
```diff
@@ -4,33 +4,100 @@
...
+        Schedule(triggers = listOf(Cron(minute = "0", hour = "0", dayWeek = "6"))), // Once a week.
     ),
     sourceFile = __FILE__,
 ) {
     job(
         id = "check",
         runsOn = RunnerType.UbuntuLatest,
     ) {
-        uses(action = Checkout())
+        uses(action = Checkout(fetchDepth = Infinite))
```
OptimumCode, line 37: note: I think it might not work. By default, as far as I know, the token used for checkout has read-only permissions. You need to request write permissions for the repo

OptimumCode: question: What do you think if instead of checking out the main branch we checkout `commits-to-upstream-badge` branch right away? `ref` parameter should allow that. In this case, you won't probably need all the staff you do with `stash` command when committing changes.

Regarding the `upstream-commit.txt` I think you can check out the main branch for SnakeKMP repo just like you checked out the `snakeyaml-engine` repo.
```
git clone --branch master --single-branch https://bitbucket.org/snakeyaml/snakeyaml-engine.git
git clone --branch ${{ github.ref }} --single-branch ${{ github.repositoryUrl }} SnakeKMP
cd snakeyaml-engine
git log --oneline $(cat ../SnakeKMP/upstream-commit.txt)..master | wc -l > ../number-of-commits.txt
```

krzema12: > I think it might not work

I tested it by triggering the workflow manually, works: https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/11724730307. I know what you mean. I tried looking up info about the default but it's somehow not straightforward.

Regarding checking out the target branch, yeah - I had it like this before, but then I used the Kotlin-based step where github-workflows-kt has to have the script checked out, hence such complication. I think having the step of building the SVG in Kotlin is not worth the hassle of later operarions on git. I'll rewrite this step to Bash and then everything should become simpler like you suggest.

OptimumCode: About rewriting step to bash: have you considered using [shields.io](https://shields.io/docs/static-badges)? In this case it would be just a single curl or wget request. This of course introduces dependency to that service but it will be much clearer than having svg defined as a string (in my opinion)

krzema12: I haven't, it's a great idea, thanks!

OptimumCode (approved):

PR #273, opened 2024-11-15
Fix identityHashCode for JS and WasmJS

krzema12:
Part of https://github.com/krzema12/snakeyaml-engine-kmp/issues/279

TODO:
* [x] Add tests that cover this piece for all targets


krzema12 (commented):

| Benchmark suite | Current: c410bdf2a293ae383c8299ca0ad9e1827d307699 | Previous: 2f27c2a1f4c0b5fbb2416a1f296f9def9e920799 | Ratio |
|-|-|-|-|
| `macosX64.LoadingTimeBenchmark.loadsOpenAiSchema ( {"openAiYamlPath":"data/issues/kmp-issue-204-OpenAI-API.yaml"} )` | `222.93126257499998` ms/op | `125.39226517317073` ms/op | `1.78` |

OptimumCode:
Hi, @krzema12 @aSemy. What do you think about implementing identity hash code using WeakMap for JS platform if the parameter is not an `object` (similar to wasmJs)? Right now, I think, this is the only thing that blocks this PR from going forward

krzema12:
Yeah, worth trying, I'm just focusing on porting other simpler changes for now. If you want to tackle it, feel welcome!

OptimumCode:
Turns out, you can only use objects or non-registered symbols as a key for WeakMap (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap/set#key). Looking for another way to make it work

krzema12:
@OptimumCode so the problematic piece is the primitive types, right? Have you considered converting them to something that can be used as a WeakMap key so that the 1:1 relation is preserved?

OptimumCode:
> Have you considered converting them to something that can be used as a WeakMap key so that the 1:1 relation is preserved?

Yes, but honestly I don't know what could be used here since creating an object would mean that each primitive value would have a unique identity. We can simply increment the internal counter we use on every primitive type instead:

```kotlin
    if (jsTypeOf(any) != "object") {
        return IdentityHashCode(lastIdentityHashCodeId++)
    }
```

I was also thinking about an alternative approach:

The main reason for having the identity map in the representer is to prevent recursion. Right now, the implementation uses `IdentityHashCode` class as a key (value type wrapping `Int`). Since the primitive types and strings cannot cause the recursion in any case, we could change what we use as a key in the map. If the key is a primitive or a string - we use the object itself as a key. Otherwise, we create a wrapper object containing an identity hashcode for the key and store this object instead. Something like this:

```kotlin

val map: MutableMap<Any, Node> = ...
val key: Any? = ...
if (isPrimitiveOrString(key)) {
  map[key] = ...
} else {
  map[IdentityHashCode(identityHashCode(key))] = ...
}
```

krzema12:
> Yes, but honestly I don't know what could be used here since creating an object would mean that each primitive value would have a unique identity.

I was thinking about dynamically building a symbol name, with gluing the primitive's type and value. I don't know how much sense it makes in practice, though.

Your idea sounds good! Definitely worth testing.

OptimumCode:
Hi, @krzema12 @aSemy. Could you please take a look at the changes when you have time?

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/SystemTest.kt
```diff
@@ -0,0 +1,39 @@
+package it.krzeminski.snakeyaml.engine.kmp.internal
+
+import io.kotest.core.spec.style.FunSpec
+import io.kotest.matchers.shouldBe
+import io.kotest.matchers.shouldNotBe
+
+class SystemTest : FunSpec({
```
krzema12, line 7: For completeness, I think it would be good to add a test for calling `identityHashCode` on an int or a string (some types that are considered primitive on all or some platforms), and to achieve it, we should make it behave the same way on all platforms.

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/representer/BaseRepresenter.kt
```diff
@@ -232,36 +231,43 @@ abstract class BaseRepresenter(
...
+        null,
+        is Byte,
+        is Short,
+        is Int,
+        is Long,
+        is Float,
+        is Double,
+        is Boolean,
+        is Char,
+        is String -> this
```
krzema12, line 265: Worth adding a comment that these types are primitive on at least one target platform (especially confusing for String).

krzema12 (commented):
Great job! Just a few nits.

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/SystemTest.kt
OptimumCode: Right now it would cause an exception for JS (since it expects only objects). We could:
- throw the same exception on all other platforms (might make method call a bit slower);
- add a documentation about how the "identity hashcode" behaves differently on some of the platforms.

To be honest, I am not sure what would be the best way here. Probably the first one but we would need to measure performance impact for additional checks. I will take a look at the first option if no objections from your side

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/representer/BaseRepresenter.kt
OptimumCode: Good point, will do

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/SystemTest.kt
krzema12: The first option was my idea. Sure, let's see what the benchmarks tell us, we have them thanks to you! 😊

OptimumCode:
Considering the error of measurement it look more or less okay
```
Before:

macosX64.DumpBenchmark.dumpMapWithAllTypes ( {} )	91.7295171144458 us/op		
mingwX64.DumpBenchmark.dumpMapWithAllTypes ( {} )	85.99261374108332 us/op		
macosArm64.DumpBenchmark.dumpMapWithAllTypes ( {} )	21.850650706950642 us/op		
jvm.DumpBenchmark.dumpMapWithAllTypes	                15.243241587097435 us/op		
js.DumpBenchmark.dumpMapWithAllTypes ( {} )            115.11791006436947 us/op		
linuxX64.DumpBenchmark.dumpMapWithAllTypes ( {} )	52.52259395065473 us/op		

After:

macosX64.DumpBenchmark.dumpMapWithAllTypes ( {} )	67.95420038861793 us/op		
mingwX64.DumpBenchmark.dumpMapWithAllTypes ( {} )	86.26141539258876 us/op		
macosArm64.DumpBenchmark.dumpMapWithAllTypes ( {} )	20.78312291351127 us/op		
jvm.DumpBenchmark.dumpMapWithAllTypes	                17.498812364098967 us/op		
js.DumpBenchmark.dumpMapWithAllTypes ( {} )             98.66989433118133 us/op		
linuxX64.DumpBenchmark.dumpMapWithAllTypes ( {} )	56.87801296996396 us/op		
```

snake-kmp-benchmarks/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/benchmark/DumpBenchmark.kt
```diff
@@ -0,0 +1,47 @@
...
+            buffer.writeUtf8(str)
+        }
+
+        override fun write(str: String, off: Int, len: Int) {
+            buffer.writeUtf8(str, off, len)
+        }
+
+    }
+
+    private val objectToDump: Any = linkedMapOf(
```
krzema12, line 30: I'm a little worried that this structure is too simple to give meaningful results in the benchmark, but it's certainly better than nothing. Thanks for adding it!

krzema12:
@OptimumCode thanks for the cooperation here! 🙇

Issue #290, opened 2024-12-14
Port "Scanner#checkToken(Token.ID) avoids varargs"


krzema12:
https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/b26779e3887d8d6b89e7e3c375bc497e3c8c5be9

PR #302, opened 2024-12-16
[Port from snakeyaml-engine] Scanner#checkToken(Token.ID) avoids varargs

krzema12:
Closes https://github.com/krzema12/snakeyaml-engine-kmp/issues/290

Ports https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/b26779e3887d8d6b89e7e3c375bc497e3c8c5be9

### DIFF
--- a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/Scanner.kt
+++ b/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/Scanner.kt
@@ -35,6 +35,18 @@ interface Scanner : Iterator<Token> {
     fun checkToken(vararg choices: Token.ID): Boolean
 
     /**
+     * Check if the next token is the given type.
+     *
+     * @param choice token ID to match with
+     * @return `true` if the next token is the given type. Returns `false`
+     * if no more tokens are available.
+     * @throws ScannerException Thrown in case of malformed input.
+     */
+    fun checkToken(choice: Token.ID): Boolean {
+        return checkToken(choices=arrayOf(choice))
+    }
+
+    /**
      * Return the next token, but do not delete it from the stream.
      *
      * @return The token that will be returned on the next call to [next]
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/Scanner.kt
```diff
@@ -34,6 +34,18 @@ interface Scanner : Iterator<Token> {
...
 
+    /**
+     * Check if the next token is the given type.
+     *
+     * @param choice token ID to match with
+     * @return `true` if the next token is the given type. Returns `false`
+     * if no more tokens are available.
+     * @throws ScannerException Thrown in case of malformed input.
+     */
+    fun checkToken(choice: Token.ID): Boolean {
```
OptimumCode, line 45: issue: this causes compilation error for Java classes implementing this interface - by default, default methods are not visible in Java. We could use compiler options described [here](https://kotlinlang.org/docs/java-to-kotlin-interop.html#default-methods-in-interfaces) but it might break binary compatibility (we need to be careful here)

OptimumCode, line 45: chore: Please, update the API dump - this is a new method in the public API

OptimumCode (changes requested):

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/Scanner.kt
krzema12: I wasn't aware of that, thanks! I'll see if migrating all Java inheritors is doable. Probably yes - I cannot check right now, but I think there was just one consumer in the tests.

krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/305

krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/302/commits/8f018c02dc6784688a2b50f916f23636488c38d6

Issue #304, opened 2024-12-17
Publishing snapshots from main branch doesn't work


krzema12:
See e. g. https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/12362611636/job/34502277861

krzema12:
@aSemy do you recall why we publish the artifacts to GitHub vs. Maven's snapshot repository? For the latter, we could reuse the logic we use for publishing the releases, and this way we could also test this logic with every push to main. Do you see some arguments against migration to this approach?

aSemy:
We added GitHub branch publishing right at the start before Maven Central publishing was configured. It was easy to configure and set up, but it's a hack. It's a workaround for GitHub Packages forbidding anonymous access.

It can definitely be removed and replaced with Maven Central snapshots.

krzema12:
Let's see if the release workflow used for releasing to Maven Central works if ran from the main branch: https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/12381982894/job/34561992660

Waiting for `krzeminski` dir in https://s01.oss.sonatype.org/content/repositories/snapshots/it/

krzema12:
It worked (wrong URL above): https://oss.sonatype.org/content/repositories/snapshots/it/krzeminski/
So the next step is to hook this workflow so that it's called for every commit on `main`.

aSemy:
I suggest we delete the `artifacts` branch.

https://github.com/krzema12/snakeyaml-engine-kmp/tree/artifacts

It takes up a significant amount of space (try downloading a zip of the repo!)

Theoretically some project could depend on it as a Maven repository, but I doubt it, and it should be easy for them to instead use Maven Central.

krzema12:
Done, thanks!

PR #308, opened 2024-12-18
ci: publish snapshot releases to Sonatype snapshots repo

krzema12:
Closes https://github.com/krzema12/snakeyaml-engine-kmp/issues/304

GitHub branch-based snapshots publishing was something we started with because it seemed to be easier to configure. Now that we have a proper workflow for publishing proper releases, it can be easily reused for publishing snapshot releases.

Tested here:
* https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/12381982894/job/34561992660
* https://oss.sonatype.org/content/repositories/snapshots/it/krzeminski/


.github/workflows/release.main.kts
```diff
@@ -35,9 +35,9 @@ val SIGNING_KEY by Contexts.secrets
 val SIGNING_PASSWORD by Contexts.secrets
 
 workflow(
-    name = "Publish release to Maven Central",
+    name = "Publish release to Maven Central or snapshot repo",
     on = listOf(
-        WorkflowDispatch()
```
OptimumCode, line 40: I would suggest keeping a manual trigger for workflow. It might be handy when you need a patch/backport in older major release. In this case, you can easily trigger publication for a different branch

OptimumCode (dismissed):

.github/workflows/release.main.kts
krzema12: Fair point! Changed: https://github.com/krzema12/snakeyaml-engine-kmp/pull/308/commits/39076249d067fec92cd3c3089c6d6af10af3aa23

OptimumCode (approved):

Issue #292, opened 2024-12-14
Port "Allow HighSurrogate to be the last char in the data window"


krzema12:
https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/c36e42f972c53d851a91e2a2c6323a4c066da662

PR #314, opened 2024-12-30
[Port from snakeyaml-engine] Allow HighSurrogate to be the last char in the data window

krzema12:
Ports https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/c36e42f972c53d851a91e2a2c6323a4c066da662
This PR actually doesn't modify the behavior because we use Okio which handles reading UTF-8 for us. We just need to port the tests and make some adjustments in the docstrings.

Closes https://github.com/krzema12/snakeyaml-engine-kmp/issues/292

### DIFF
--- a/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
+++ b/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
@@ -36,7 +36,7 @@ class ReaderStringTest : FunSpec({
             } else if (i.toChar().isLowSurrogate()) {
                 lowSurrogatesCounter++
             } else {
-                val str = CharArray(1) { i.toChar() }.concatToString()
+                val str = charArrayOf(i.toChar()).concatToString()
                 val regularExpressionResult = StreamReader.isPrintable(str)
 
                 var charsArrayResult = true
@@ -65,7 +65,7 @@ class ReaderStringTest : FunSpec({
 
     test("high surrogates alone") {
         for (i in 0xD800..0xDBFF) {
-            val str = CharArray(1) { i.toChar() }.concatToString()
+            val str = charArrayOf(i.toChar()).concatToString()
             val reader = StreamReader(LoadSettings.builder().build(), str)
             reader.peek() shouldBe '?'.code
         }
@@ -73,7 +73,7 @@ class ReaderStringTest : FunSpec({
 
     test("low surrogates alone") {
         for (i in 0xDC00..0xDFFF) {
-            val str = CharArray(1) { i.toChar() }.concatToString()
+            val str = charArrayOf(i.toChar()).concatToString()
             val reader = StreamReader(LoadSettings.builder().build(), str)
             reader.peek() shouldBe '?'.code
         }
--- a/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
+++ b/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
@@ -41,7 +41,7 @@ class ReaderStringTest : FunSpec({
                     StreamReader(LoadSettings.builder().build(), str).peek()
                 } catch (e: Exception) {
                     (
-                        e.message?.startsWith("unacceptable character") ?: false
+                        (e.message?.startsWith("unacceptable character") ?: false)
                             || e.message == "special characters are not allowed"
                     ) shouldBe true
                     charsArrayResult = false
@@ -59,13 +59,13 @@ class ReaderStringTest : FunSpec({
 
     test("high surrogate alone") {
         val reader = StreamReader(LoadSettings.builder().build(), "test\uD800")
-        shouldThrow<ReaderException> {
+        try {
             while (reader.peek() != 0) {
                 reader.forward()
             }
-        }.also {
-            it.toString() shouldContain "(0xD800) The last char is HighSurrogate (no LowSurrogate detected)"
-            it.position shouldBe 5
+        } catch(e: ReaderException) {
+            e.toString() shouldContain "(0xD800) The last char is HighSurrogate (no LowSurrogate detected)"
+            e.position shouldBe 5
         }
     }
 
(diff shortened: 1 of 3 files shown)


krzema12:
The tests fail because for character with code point 56320, one of the tests that checks human-readability concludes that two methods of checking it return different results.

krzema12:
It _may_ be somehow related to something that we perhaps didn't manage to port properly:

https://github.com/krzema12/snakeyaml-engine-kmp/blob/509b5117a0abcf92d96f8fedde673615dcf6d21a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt#L194-L201

krzema12:
It turned out that the test would fail for 1024 characters: from 0xDC00 to 0xDFFF. See https://github.com/krzema12/snakeyaml-engine-kmp/blob/509b5117a0abcf92d96f8fedde673615dcf6d21a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt#L274-L282
If we added `|| c in 0xDC00 .. 0xDFFF` there, the test would pass. Now we need to clarify the discrepancy on the other side.

krzema12:
Per https://en.wikipedia.org/wiki/Unicode_block, `U+DC00..U+DFFF` is a block of low surrogates.

Other interesting and perhaps pieces of code I found:

1. https://github.com/krzema12/snakeyaml-engine-kmp/blob/509b5117a0abcf92d96f8fedde673615dcf6d21a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/utils/Character.kt#L74-L90 - why is it commented out? It seems to deal with the surrogates.
2. https://github.com/krzema12/snakeyaml-engine-kmp/blob/509b5117a0abcf92d96f8fedde673615dcf6d21a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/utils/CharSequenceExtensions.kt#L122-L134 - perhaps we need to use this instead of Kotlin's `.code` in some place? But then, why?

krzema12:
@aSemy do you happen to have some context here? Porting this change seems to be particularly hairy, there are some pieces of code that are commented out, also see my comments for partial diagnosis.

If we don't manage to make any sense of it, the next step is to take snakeyaml-engine and carefully track what happens in the unit test in question for chars in range 0xDC00-0xDFFF.

krzema12:
In snakeyaml-engine-kmp, one of the tested low surrogates is converted to `?` (a question mark), while in snakeyaml-engine it's a proper two-byte character:

Screenshot 2024-12-31 at 12 34 09
Screenshot 2024-12-31 at 12 34 20

krzema12:
Here's an important clue:

Screenshot 2024-12-31 at 12 47 54

Notice that the character in question is replaced with `?` after some transformation we currently have.

`encodeToByteArray` has in its docs:
> Any malformed char sequence is replaced by the replacement byte sequence.

krzema12:
`String.encodeToByteArray()` delegates on the JVM to https://docs.oracle.com/javase/8/docs/api/java/lang/String.html#getBytes-java.nio.charset.Charset-. Its docs say:
> This method always replaces malformed-input and unmappable-character sequences with this charset's default replacement byte array. The [CharsetEncoder](https://docs.oracle.com/javase/8/docs/api/java/nio/charset/CharsetEncoder.html) class should be used when more control over the encoding process is required.

We should dig deeper into what characters are replaced by it.

krzema12:
https://github.com/openjdk/jdk/blob/a77ed30fcc3360cd16a11b1899f52f7e871df1df/src/java.base/share/classes/java/lang/String.java#L1854-L1857

krzema12:
Here's a piece of JDK's code that I found, using the debugger, takes part in encoding the string to bytes:
https://github.com/openjdk-mirror/jdk7u-jdk/blob/f4d80957e89a19a29bb9f9807d2a28351ed7f7df/src/share/classes/java/net/URLEncoder.java#L254-L262

```java
                            if (d >= 0xDC00 && d <= 0xDFFF) {
                                /*
                                  System.out.println("\t"
                                  + Integer.toHexString(d)
                                  + " is low surrogate");
                                */
                                charArrayWriter.write(d);
                                i++;
                            }
```

krzema12:
This case is really confusing. Looks like our ported version behaves differently for low surrogates. Not sure if it's worth turning our implementation upside down just to match snakeyaml-engine's behavior for characters that anyway look like invalid UTF-8 (as far as I understand the surrogates).

I'm opting for not blocking the release on porting this particular change. Let me wait a few more days for @OptimumCode to see if you have some insights from the top of your head. If no new ideas appear, let's just keep this PR as a draft and get back to it later.

OptimumCode:
Hi, @krzema12. Since SnakeKMP reads UTF-8 strings from the `source` (in `StreamReader` class), getting an invalid UTF-8 string from it is impossible (all invalid malformed bytes are replaced). Therefore, we probably should exclude low surrogates from the range of codepoints in this test (0xDC00-0xDFFF as you mentioned before). We could add them to the test the same way the high surrogates are added.

Also, I think we should add a comment to the `StreamReader` saying that if the `source` contains invalid UTF-8 encoded byte sequences those bytes will be replaced with `?`.

OptimumCode (commented):
suggestion: I think we should also create a test for high/low surrogates. They should be replaced with a replacement character (`?` in our case)

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
```diff
@@ -0,0 +1,100 @@
...
+        val reader = StreamReader(LoadSettings.builder().build(), "test\uD800")
+        try {
+            while (reader.peek() != 0) {
+                reader.forward()
+            }
+        } catch(e: ReaderException) {
+            e.toString() shouldContain "(0xD800) The last char is HighSurrogate (no LowSurrogate detected)"
+            e.position shouldBe 5
+        }
+    }
```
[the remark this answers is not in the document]
krzema12: Right! Good catch

krzema12:
@OptimumCode great feedback, thanks! :bow: We'll manage to tick off porting this (as porting just the tests, no changes needed in the impl 🎉) for the upcoming release. Super cool!

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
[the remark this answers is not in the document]
krzema12: I'll replace it with the tests that asserts returning the replacement character "?".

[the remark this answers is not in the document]
krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/314/commits/4e25ff352d3158c11ddacba0951ab637b32de025

OptimumCode (approved):

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
```diff
@@ -0,0 +1,110 @@
...
+        for (i in 0xD800..0xDBFF) {
+            val str = CharArray(1) { i.toChar() }.concatToString()
+            val reader = StreamReader(LoadSettings.builder().build(), str)
+            reader.peek() shouldBe '?'.code
+        }
+    }
+
+    test("low surrogates alone") {
+        for (i in 0xDC00..0xDFFF) {
+            val str = CharArray(1) { i.toChar() }.concatToString()
```
OptimumCode, line 76: suggestion: `CharArray(1) {}` can be replaced with `charArrayOf(...)`

OptimumCode (dismissed):

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
OptimumCode: I just noticed that you have a requirement to have all conversations resolved - please, feel free to resolve any comments that are not marked as issue😊

krzema12: Thanks, let me apply this!

krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/314/commits/5177cda7d3bb271ea93327b5c36b0da96bfca173

OptimumCode (approved):

PR #324, opened 2025-01-03
Handle non-existing common test resources gracefully

krzema12:
Closes https://github.com/krzema12/snakeyaml-engine-kmp/issues/323

Replaces NPE with an IllegalArgumentException with a nice message.

PR #325, opened 2025-01-03
Publish build scans in case of failures

krzema12:
Helpful with troubleshooting failed builds, especially if they failed on a runner that one doesn't have available.

A direct motivation for it are failing tests in https://github.com/krzema12/snakeyaml-engine-kmp/pull/324. However, the issue is weird there: if I enable build scans in that PR, the tests pass. Adding the build scans in general is a good thing, so here's a separate PR.

### DIFF
--- a/settings.gradle.kts
+++ b/settings.gradle.kts
@@ -55,6 +55,17 @@ dependencyResolutionManagement {
 
 plugins {
     id("org.gradle.toolchains.foojay-resolver-convention") version("0.9.0")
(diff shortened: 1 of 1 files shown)


OptimumCode (approved):
All looks good to me!
Just a small note in case you would want to publish report for every build in the future: you can configure that without making changes in settings.gradle.kts. There are parameters in the gradle action (https://github.com/gradle/actions/blob/main/setup-gradle/action.yml#L128). But it is not possible to configure publication for failed builds only

Issue #269, opened 2024-11-11
Port tests to common Kotlin


krzema12:
https://github.com/krzema12/snakeyaml-engine-kmp/pull/410To be able to run the tests for all supported platforms, let's first convert them from Java to Kotlin.

Focusing on tests that use some files from the resources, starting from the tests that have most LoC:

* [X] 138 org/snakeyaml/engine/usecases/env/EnvVariableTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/332
* [X] 92 org/snakeyaml/engine/usecases/indentation/IndentWithIndicatorTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/338
* [x] 78 org/snakeyaml/engine/issues/issue39/EmitCommentAndSpacesTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/339
* [x] 71 org/snakeyaml/engine/v2/api/LoadSequenceTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/346
* [x] 71 org/snakeyaml/engine/issues/issue149/GlobalDirectivesTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/347
* [x] 70 org/snakeyaml/engine/v2/api/LoadMappingTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/348
* [x] 53 org/snakeyaml/engine/usecases/recursive/RecursiveSetTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/351
* [x] 38 org/snakeyaml/engine/v2/constructor/StandardConstructorTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/352
* [x] 38 org/snakeyaml/engine/issues/issue512/ListWithCommentTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/353
* [x] 64 ./org/snakeyaml/engine/v2/api/OptionalMarksTest.java - https://github.com/krzema12/snakeyaml-engine-kmp/pull/371
* [x] remove `TestUtils.getResource` - https://github.com/krzema12/snakeyaml-engine-kmp/pull/366

Harder ones:
* [x] 114 org/snakeyaml/engine/usecases/inherited/InheritedImportTest.java - a bigger one, uses Java's `File` class and inheritance
  * https://github.com/krzema12/snakeyaml-engine-kmp/pull/358
[...]

PR #329, opened 2025-01-03
Commonize EmitterWithCommentEnabledTest

krzema12:
Part of https://github.com/krzema12/snakeyaml-engine-kmp/issues/269

### DIFF
--- a/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/StringOutputStream.kt
+++ b/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/StringOutputStream.kt
@@ -13,7 +13,7 @@ internal class StringOutputStream(
     }
 
     override fun write(str: String, off: Int, len: Int) {
-        builder.append(str, off, len)
+        builder.append(str.drop(off).take(len))
     }
 
     override fun toString(): String = builder.toString()
--- a/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/EmitterWithCommentEnabledTest.kt
+++ b/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/EmitterWithCommentEnabledTest.kt
@@ -135,7 +135,7 @@ class EmitterWithCommentEnabledTest : FunSpec({
 
         val result = runEmitterWithCommentsEnabled(data)
         // We currently strip Directive comments
-        result shouldBe data
+        result shouldBe ""
     }
 
     test("sequence") {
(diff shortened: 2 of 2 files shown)


krzema12 (commented):

| Benchmark suite | Current: eba55ed710cd6ec58c8da33dd5abb2f2f8b30acd | Previous: fbbab6238e854a433b89b038fd6e6010ec805d83 | Ratio |
|-|-|-|-|
| `macosX64.DumpBenchmark.dumpMapWithAllTypes ( {} )` | `123.73274097997721` us/op | `55.06085171989266` us/op | `2.25` |
| `macosX64.LoadingTimeBenchmark.loadsOpenAiSchema ( {"openAiYamlPath":"data/issues/kmp-issue-204-OpenAI-API.yaml"} )` | `213.66538299473686` ms/op | `107.36575578695651` ms/op | `1.99` |

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/StringOutputStream.kt
```diff
@@ -13,7 +13,7 @@ internal class StringOutputStream(
     }
 
     override fun write(str: String, off: Int, len: Int) {
-        builder.append(str, off, len)
+        builder.append(str.drop(off).take(len))
```
krzema12, line 16: While here, fixing a bug, which is necessary to make the ported tests work. It should be
```
builder.append(str, off, off + len)
```
because the last arg is an end offset, not length. I wanted to use named arguments here, but I'm getting

```
e: file:///Users/piotr/repos/snakeyaml-engine-kmp/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/StringOutputStream.kt:16:17 None of the following candidates is applicable:
fun append(p0: Any!): StringBuilder!
fun append(p0: String!): StringBuilder!
fun append(p0: StringBuffer!): StringBuilder!
fun append(p0: CharSequence!): StringBuilder!
fun append(p0: CharSequence!, p1: Int, p2: Int): StringBuilder!
fun append(p0: CharArray!): StringBuilder!
fun append(p0: CharArray!, p1: Int, p2: Int): StringBuilder!
fun append(p0: Boolean): StringBuilder!
fun append(p0: Char): StringBuilder!
fun append(p0: Int): StringBuilder!
fun append(p0: Long): StringBuilder!
fun append(p0: Float): StringBuilder!
fun append(p0: Double): StringBuilder!
```

which is pretty interesting, looks like a bug in Kotlin. Even IntelliJ places the named arguments and then compilation fails. That's why I decided to not use the overload with indices, and select the substring with `drop` and `take`.

krzema12: Reported as https://youtrack.jetbrains.com/issue/KT-74222/

OptimumCode (approved):

Issue #266, opened 2024-11-09
Investigate worse performance on JVM compared to snakeyaml-engine


krzema12:
See:
![Screenshot_2024-11-09-11-22-03-910_com android chrome-edit](https://github.com/user-attachments/assets/d6bf6e5c-bcf6-4bfe-83b4-75c7c18a9f44)
![Screenshot_2024-11-09-11-21-30-106_com android chrome-edit](https://github.com/user-attachments/assets/102c021d-23f4-4824-8f1f-9bc7c358b106)

The KMP version is about 2x slower. It's likely a regression when porting to Kotlin, e. g. we went with some immutable structure that results in copying data on each update.

Let's understand this difference, perhaps by using a profiler.

OptimumCode:
I did JFR recordings of the benchmark for SnakeKMP and SnakeYAML.

This is the image for SnakeYAML run. All looks good (note that there is almost nothing related to work with the source)
![image](https://github.com/user-attachments/assets/ee49f854-7b02-41d1-aa47-e6743c6fda48)

And this is the image for SnakeKMP run. We can see that significant time is spent on decoding UTF-8 strings
![image](https://github.com/user-attachments/assets/0c099d76-9d8c-4113-b0fd-6f85c5f93ad9)

The remaining part of the graph looks more or less the same. So, I think, to make the performance close to SnameYAML we should invest some time in refactoring the part that reads data from the source

OptimumCode:
The benchmark results
[bench.tar.gz](https://github.com/user-attachments/files/18302094/bench.tar.gz)

OptimumCode:
Also, most likely, the long time decoding UTF-8 is related to the fact that we use the `readUtf8()` method. This method will read **all bytes** from the source and decode them into a string. This means we read the whole file at once. Maybe, we could try to migrate to `readUtf8Line` method (at leas that would limit the number of bytes we read)

krzema12:
@OptimumCode Great findings, thanks! Do you feel like pushing it forward, so experimenting with some optimizations? It doesn't have to happen soon, in a month or a quarter is fine 😄 Porting the tests to the common source set will keep me busy for a while.

OptimumCode:
I will try to take a look at what can be done there

PR #331, opened 2025-01-04
performance: avoid making codepoints window copy on prefix call

OptimumCode:
I noticed that we make a copy of a codepoint window when `prefix(int)` method is called. It looks redundant since the `StreamReader` should not be invoked from different threads and the resulting `String` is immutable.

I have changed the code build string without making a copy and got a significant performance increase on local benchmarks (let see how it behaves in CI)

```text
Benchmark                                                 Mode  Cnt   Score   Error  Units
LoadingTimeBenchmark.loadsOpenAiSchema                    avgt   10  15.655 ± 0.138  ms/op
SnakeyamlEngineJvmLoadingTimeBenchmark.loadsOpenAiSchema  avgt   10  13.554 ± 0.348  ms/op
```

Related to #266


krzema12:
Indeed there's some improvement, 30 -> 26 ms/op for the JVM, and a stunning 200 -> 120 ms/op for JS! Let's go ahead with this, great job!

krzema12 (approved):

PR #332, opened 2025-01-04
Kotlinize and partially commonize EnvVariableTest

krzema12:
Part of https://github.com/krzema12/snakeyaml-engine-kmp/issues/269


src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/usecases/env/EnvVariableTest.kt
```diff
@@ -0,0 +1,79 @@
...
+import io.kotest.matchers.string.shouldEndWith
+import it.krzeminski.snakeyaml.engine.kmp.api.Load
+import it.krzeminski.snakeyaml.engine.kmp.api.LoadSettings
+import it.krzeminski.snakeyaml.engine.kmp.env.EnvConfig
+import it.krzeminski.snakeyaml.engine.kmp.exceptions.MissingEnvironmentVariableException
+import it.krzeminski.snakeyaml.engine.kmp.stringFromResources
+
+class EnvVariableTest : FunSpec({
+    // the variable is set by Gradle
+    val VALUE1 = "EnvironmentValue1"
```
krzema12, line 15: TODO: check the support for reading env vars for targets other than the JVM, currently the tests fail for these.

OptimumCode: Kotest has a method to read env variables. Probably, we could copy that part from them:

https://github.com/kotest/kotest/blob/master/kotest-common/src/commonMain/kotlin/io/kotest/mpp/sysprop.kt

OptimumCode: There is also a library that implements that already:

https://klibs.io/project/luca992/getenv-kt

krzema12: FTR, for JS we now have https://github.com/krzema12/snakeyaml-engine-kmp/blob/85a31ae35427bee23ec2801718d3143b55eaa4ac/src/jsCommonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/system.js.kt#L3-L5
which explains why it doesn't work despite setting the vars in Gradle, and for native, there's a chance it will work:
https://github.com/krzema12/snakeyaml-engine-kmp/blob/85a31ae35427bee23ec2801718d3143b55eaa4ac/src/nativeMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/system.native.kt#L10-L12

krzema12: The impl of https://klibs.io/project/luca992/getenv-kt is so simple I'll probably just copy it over to our lib, thanks for finding it!

OptimumCode: It is simple for all platforms except for `wasmWasi`. If/when this platform is supported by SnakeKMP we probably should use the `getenv-kt` instead

krzema12: Hmm, right. I'd like to avoid adding an extra dependency for now (various consequences), but you're right that in the future we may want to revisit.

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/usecases/env/EnvVariableTest.kt
```diff
@@ -0,0 +1,80 @@
...
+    val NULL_ENV_CONFIG = object : EnvConfig {
+        override fun getValueFor(name: String, separator: String?, value: String?, environment: String?): String? {
+            return null
+        }
+    }
+
+    test("Parse docker-compose.yaml example") {
```
krzema12, line 23: TODO: find a way to skip these for JS browser. Asked here: https://kotlinlang.slack.com/archives/CT0G9SD7Z/p1736115537677279

krzema12: Added an ugly try-catch.

build.gradle.kts
```diff
@@ -57,6 +59,16 @@ tasks.withType<Test>().configureEach {
...
 
+tasks.withType<KotlinJsTest>().configureEach {
+    environment("EnvironmentKey1", "EnvironmentValue1")
+    environment("EnvironmentEmpty", "")
+}
+
+tasks.withType<KotlinNativeTest>().configureEach {
+    environment("EnvironmentKey1", "EnvironmentValue1")
+    environment("EnvironmentEmpty", "")
+}
```
krzema12, line 70: Doesn't seem to work.

OptimumCode: I also had some issues with setting env variables that are visible tests. Maybe this will help:

https://github.com/OptimumCode/json-schema-validator/blob/main/test-suites/build.gradle.kts#L137

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/usecases/env/EnvVariableTest.kt
OptimumCode: I think we can create a function like this:

Common:
```kotlin
expect fun isEnvUnsupported(): Boolean
```
js:
```kotlin
// returns 'true' is we are in the browser
actual fun isEnvUnsupported(): Boolean = js("return (typeof window !== 'undefined' && typeof document !== 'undefined')")
```

nonJs:
```kotlin
actual fun isEnvUnsupported(): Boolean = false
```

And this can be used to disable env tests in browser

src/wasmJsMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/system.wasm.kt
```diff
@@ -1,5 +1,9 @@
 package it.krzeminski.snakeyaml.engine.kmp.internal
 
+internal actual fun getEnvironmentVariable(key: String): String? {
+    TODO()
+}
```
OptimumCode, line 5: issue: this causes `Conflicting overloads` error - we have same method in `jsCommon` sources

src/jsCommonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/process.kt
```diff
@@ -0,0 +1,7 @@
+package it.krzeminski.snakeyaml.engine.kmp.internal
+
+public external val process: Process
+
+public external interface Process {
+    public val env: dynamic
+}
```
OptimumCode, line 7: issue: This does not work for `wasmJs` but `wasmJs` depends on `jsCommon` sources. And this causes compilation errors. We should move it to `jsMain` sources

krzema12: Thanks, I couldn't understand what jsCommonMain means, now I see it's shared between regular and WASM JS!

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/usecases/env/EnvVariableTest.kt
krzema12: Good one, I was about to do something similar.
https://github.com/krzema12/snakeyaml-engine-kmp/pull/332/commits/569555d43743977b05ab1ee3b23421670576b031

src/jsTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/TargetUtils.kt
```diff
@@ -0,0 +1,5 @@
+package it.krzeminski.snakeyaml.engine.kmp.internal
+
+// returns 'true' is we are in the browser
+actual fun areEnvVarsSupported(): Boolean =
+    js("return (typeof window === 'undefined' || typeof document === 'undefined')") as Boolean
```
krzema12, line 5: Double-check it returns true for Nodejs.

krzema12: It does:
Screenshot_2025-01-06-21-41-53-066_com android chrome-edit

build.gradle.kts
krzema12: Oh, there's a separate config for simulator-based tests... Thanks, will try it out!

krzema12: Works, thanks!

src/jsTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/TargetUtils.kt
```diff
@@ -0,0 +1,5 @@
+package it.krzeminski.snakeyaml.engine.kmp.internal
+
+// returns 'true' is we are in the browser
```
OptimumCode, line 3: note: probably initially it was true initially but right now it returns `true` if we are not in the browser

krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/332/commits/76fc1145106324faff2f934812a2e2bd75aa5f67

krzema12 (commented):

| Benchmark suite | Current: 76fc1145106324faff2f934812a2e2bd75aa5f67 | Previous: f9c093398fe0ae732ac6863e0d0bf737d191ad4e | Ratio |
|-|-|-|-|
| `macosX64.DumpBenchmark.dumpMapWithAllTypes ( {} )` | `105.43475320362373` us/op | `55.697519886529314` us/op | `1.89` |
| `macosX64.LoadingTimeBenchmark.loadsOpenAiSchema ( {"openAiYamlPath":"data/issues/kmp-issue-204-OpenAI-API.yaml"} )` | `187.71732909545455` ms/op | `87.26484693050848` ms/op | `2.15` |

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/TargetUtils.kt
```diff
@@ -0,0 +1,3 @@
+package it.krzeminski.snakeyaml.engine.kmp.internal
+
+expect fun areEnvVarsSupported(): Boolean
```
OptimumCode, line 3: suggestion: the name of the method says pretty much all but I think it would be good to add a comment why we need it (because env variables are not available in the browser)

OptimumCode (approved):

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/TargetUtils.kt
krzema12: This is the `expect` declaration, and as such I'd refrain from leaking any impl details of concrete targets. E.g. with mentioning the browser we'd leak details about JS. It's true that we created this function solely to address no env vars in the browser, but in general we should keep things abstract. I can add some high-level comment here.

PR #335, opened 2025-01-07
Fix reading env vars for JS and WASM JS (only NodeJS)

krzema12:
While working on https://github.com/krzema12/snakeyaml-engine-kmp/pull/332, I found out that the implementations for JS and WASM JS are missing. This PR focuses on adding them. I don't want to add these in the PR that is related solely to kotlinizing and commonizing the tests to give this fix a proper visibility in the release notes.

This change was tested within https://github.com/krzema12/snakeyaml-engine-kmp/pull/332, and the goal is to merge that PR soon after this one lands.

Note: env vars work only in a non-browser environment.

### DIFF
--- a/src/jsMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/process.kt
+++ b/src/jsMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/process.kt
@@ -1,0 +1,7 @@
+package it.krzeminski.snakeyaml.engine.kmp.internal
+
+external val process: Process
+
+external interface Process {
+    val env: dynamic
+}
--- a/src/jsMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/process.kt
+++ b/src/jsMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/process.kt
@@ -1,7 +1,0 @@
-package it.krzeminski.snakeyaml.engine.kmp.internal
-
-external val process: Process
-
-external interface Process {
-    val env: dynamic
-}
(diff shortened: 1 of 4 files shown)


src/jsMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/process.kt
```diff
@@ -0,0 +1,7 @@
+package it.krzeminski.snakeyaml.engine.kmp.internal
+
+external val process: Process
+
+external interface Process {
```
OptimumCode, line 5: question: is it possible to use `internal` modifier here? Right now they are public and, I believe, will be able for everyone (I don't think we want that)

OptimumCode (dismissed):

src/jsMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/process.kt
krzema12: Sure, I moved them to system.js.kt, like we have with other `external` entities, just because I find it confusing to have `internal external` pair of modifiers 😆

krzema12 (commented):

| Benchmark suite | Current: 185652b86b0b76df056e8fd72df6d28e025be3e6 | Previous: f9c093398fe0ae732ac6863e0d0bf737d191ad4e | Ratio |
|-|-|-|-|
| `macosX64.LoadingTimeBenchmark.loadsOpenAiSchema ( {"openAiYamlPath":"data/issues/kmp-issue-204-OpenAI-API.yaml"} )` | `132.15177967894738` ms/op | `87.26484693050848` ms/op | `1.51` |

OptimumCode (approved):

PR #336, opened 2025-01-07
Re-enable WASM JS tests

### LINKED ISSUE #320: Wasm tests fail on Kotlin 2.1.0 and/or Kotest 6.0
See https://github.com/krzema12/snakeyaml-engine-kmp/pull/231

We want to unblock releasing the newest version of the lib, and since Wasm is experimental in general, we'll skip these tests. A related discussion is happening in https://kotlinlang.slack.com/archives/CDFP59223/p1734964013996149 where we hope to understand the problem better. This issue is here to remember about revisiting this problem.

krzema12:
Closes #320.

Related to https://kotlinlang.slack.com/archives/CDFP59223/p1736171757249559?thread_ts=1734964013.996149&cid=CDFP59223

> The problem in linkage of coroutines 1.8.0 (which was built against Kotlin/Wasm 1.9.21) with Kotlin stdlib 2.1.0.
In 2.1.0 we have some changes in our linkage process (including changes in stdlib), so I’d recommend to use new Kotlin coroutines which was built with Kotlin 2.1.0
For example the latest one with version 1.10.1

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -1,6 +1,4 @@
 import buildsrc.utils.configureGradleDaemonJvm
-import org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootExtension
-import org.jetbrains.kotlin.gradle.targets.js.nodejs.NodeJsRootPlugin
 
 plugins {
     buildsrc.conventions.lang.`kotlin-multiplatform`
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -24,10 +24,6 @@ kotlin {
             dependencies {
                 implementation("com.squareup.okio:okio:3.9.1")
                 implementation("net.thauvin.erik.urlencoder:urlencoder-lib:1.6.0")
-                // Overridig coroutines' version to solve a problem with WASM JS tests.
-                // See https://kotlinlang.slack.com/archives/CDFP59223/p1736191408326039?thread_ts=1734964013.996149&cid=CDFP59223
-                // TODO: remove this workaround in https://github.com/krzema12/snakeyaml-engine-kmp/issues/337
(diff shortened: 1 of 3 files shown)


build.gradle.kts
```diff
@@ -24,6 +24,10 @@ kotlin {
             dependencies {
                 implementation("com.squareup.okio:okio:3.9.1")
                 implementation("net.thauvin.erik.urlencoder:urlencoder-lib:1.6.0")
+                // Overridig coroutines' version to solve a problem with WASM JS tests.
+                // See https://kotlinlang.slack.com/archives/CDFP59223/p1736191408326039?thread_ts=1734964013.996149&cid=CDFP59223
+                // TODO: remove this workaround in https://github.com/krzema12/snakeyaml-engine-kmp/issues/337
+                runtimeOnly("org.jetbrains.kotlinx:kotlinx-coroutines-core:1.10.1")
```
OptimumCode, line 30: issue: we should add this dependency to the `commonTest` source set because we don't use them in the main library

OptimumCode (changes requested):

build.gradle.kts
krzema12: Oops, right! Moved.

krzema12 (commented):

| Benchmark suite | Current: 1a94720c4d6377567fd55a1673c9bdbd78dbddc1 | Previous: 5b8059f27f9614a881e899d726b3c86d7faad6b8 | Ratio |
|-|-|-|-|
| `macosX64.LoadingTimeBenchmark.loadsOpenAiSchema ( {"openAiYamlPath":"data/issues/kmp-issue-204-OpenAI-API.yaml"} )` | `148.95564078387096` ms/op | `83.77391314166667` ms/op | `1.78` |

Issue #354, opened 2025-01-10
Generating multiplatform resource accessor doesn't scale well


krzema12:
I'm about to port tests relying on test resources from `inherited_yaml_1_1` directory, but just after moving this dir to the common resources, I got:

```
exception: java.lang.RuntimeException: Error generating class file CommonTestResources.class (compiled from [/Users/piotr/repos/snakeyaml-engine-kmp/build/generated-code-for-resources/src/commonTest/kotlin/CommonTestResources.kt]): Method too large: CommonTestResources.<clinit> ()V
e
```

so we're hitting a limitation of Java.

krzema12:
My first idea is to naively have a separate method for initializing every directory, but then there's a question if there's any other limit (e.g. on the class size).

krzema12:
Good enough for now after the last fix.

PR #355, opened 2025-01-10
Allow larger common test resources

krzema12:
Part of https://github.com/krzema12/snakeyaml-engine-kmp/issues/354

Before this change, the build logic took all test resources, and converted them into Kotlin code that was building a map using a single expression with using nested `mapOf` function to represent the directory structure. However, when trying to move existing JVM-specific resources (there's a lot of them), one could see a failing build with:

```
exception: java.lang.RuntimeException: Error generating class file CommonTestResources.class (compiled from [/Users/piotr/repos/snakeyaml-engine-kmp/build/generated-code-for-resources/src/commonTest/kotlin/CommonTestResources.kt]): Method too large: CommonTestResources.<clinit> ()V
e
```

This change puts every call to `mapOf(...)` or `ByteString.of(...)` into a separate method, so that now the limitation is either the maximum number of files/directories in a given directory, or size of a single file. It should unblock migrating tests to the common source set that rely on the large number of files (see a work-in-progress PR: https://github.com/krzema12/snakeyaml-engine-kmp/pull/356).

It was still faster to not use KotlinPoet, but the generated code became even more ugly. It's likely that I'm going to use KotlinPoet, especially that I'm planning to extract this functionality into a library.

### DIFF
--- a/buildSrc/src/main/kotlin/buildsrc/tasks/ConvertCommonTestResourcesToKotlin.kt
+++ b/buildSrc/src/main/kotlin/buildsrc/tasks/ConvertCommonTestResourcesToKotlin.kt
@@ -74,7 +74,7 @@ abstract class ConvertCommonTestResourcesToKotlin @Inject constructor(
     }
 
     private fun generateFunctions(map: Map<String, Any>, stringBuilder: StringBuilder, path: String = "") {
-        stringBuilder.appendLine(generateSingleFunction(map, path))
+        stringBuilder.generateSingleFunction(map, path)
         stringBuilder.appendLine()
         for ((key, value) in map) {
             when (value) {
@@ -89,12 +89,13 @@ abstract class ConvertCommonTestResourcesToKotlin @Inject constructor(
         }
     }
 
-    private fun generateSingleFunction(map: Map<String, Any>, path: String): String {
+    private fun StringBuilder.generateSingleFunction(map: Map<String, Any>, path: String) {
         val functionName = getFunctionName(path)
-        return map.entries.joinToString(
+        map.entries.joinTo(
+            this,
             separator = ",\n",
             prefix = "fun ${functionName}() = mapOf(\n",
-            postfix = "\n)",
+            postfix = "\n)\n",
         ) { (key, _) ->
             "\"$key\" to ${getFunctionName("$path/$key")}()"
         }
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/kotlin/buildsrc/tasks/ConvertCommonTestResourcesToKotlin.kt
```diff
@@ -59,29 +59,48 @@ abstract class ConvertCommonTestResourcesToKotlin @Inject constructor(
...
+            postfix = "\n)",
+        ) { (key, _) ->
+            "\"$key\" to ${getFunctionName("$path/$key")}()"
+        }
+    }
+
+    private val pathToFunctionName: MutableMap<String, String> = mutableMapOf()
+
+    private fun getFunctionName(path: String): String =
+        pathToFunctionName.getOrPut(path) { "function${pathToFunctionName.size}" }
```
krzema12, line 105: Generating unique function names, exactly one name for each path in the file tree.

buildSrc/src/main/kotlin/buildsrc/tasks/ConvertCommonTestResourcesToKotlin.kt
```diff
@@ -59,29 +59,48 @@ abstract class ConvertCommonTestResourcesToKotlin @Inject constructor(
...
+                    @Suppress("UNCHECKED_CAST")
+                    generateFunctions(map[key] as Map<String, Any>, stringBuilder, "$path/$key")
+                }
             }
         }
     }
+
+    private fun generateSingleFunction(map: Map<String, Any>, path: String): String {
+        val functionName = getFunctionName(path)
+        return map.entries.joinToString(
```
OptimumCode, line 93: suggestion: You could pass the `StringBuilder` as a parameter to the function (or as a receiver) and use `joinTo` method instead of `joinToString`

OptimumCode (dismissed):

buildSrc/src/main/kotlin/buildsrc/tasks/ConvertCommonTestResourcesToKotlin.kt
krzema12: Makes sense, TIL (haven't known `joinTo`).

OptimumCode (approved):

Issue #307, opened 2024-12-17
Building on MacOS is flaky


krzema12:
Sometimes, more than less often, build-on-MacOSLatest fails.

Example: https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/12377077417/job/34545763964

From the logs:

```
  :jsBrowserTest: java.lang.IllegalStateException: Errors occurred during launch of browser for testing.
For more on this, please refer to https://docs.gradle.org/8.11.1/userguide/command_line_interface.html#sec:command_line_warnings in the Gradle documentation.
  - ChromeHeadless
  Please make sure that you have installed browsers.
  Or change it via
  browser {
      testTask {
          useKarma {
              useFirefox()
              useChrome()
              useSafari()
          }
      }
  }
```

Several commits earlier, it succeeded: https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/12369132747/job/34520445069

krzema12:
Chrome is available on the runner: https://github.com/actions/runner-images/blob/macos-14-arm64/20241211.616/images/macos/macos-14-arm64-Readme.md#browsers

krzema12:
We had just a few flakes so far. I'd wait a week or so to see how it behaves.

krzema12:
The issue seems to be gone, no sign of it in the last 2 weeks.

OptimumCode:
It looks like it reproduced again:
https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/12687697438/job/35362699974

krzema12:
Checking the included software on the worker: https://github.com/actions/runner-images/blob/macos-14-arm64/20250106.725/images/macos/macos-14-arm64-Readme.md

### Browsers
- Safari 18.2 (19620.1.16.111.6)
- SafariDriver 18.2 (19620.1.16.111.6)
- Google Chrome 131.0.6778.205
- Google Chrome for Testing 131.0.6778.204
- ChromeDriver 131.0.6778.204
- Selenium server 4.27.0

#### Environment variables
| Name            | Value                                   |
| --------------- | --------------------------------------- |
| CHROMEWEBDRIVER | /usr/local/share/chromedriver-mac-arm64 |
| EDGEWEBDRIVER   |                                         |
| GECKOWEBDRIVER  |                                         |

krzema12:
[...]

PR #367, opened 2025-01-17
Increase browserSocketTimeout parameter for karma tests

OptimumCode:
There is one more parameter that can help reduce the chances of getting error:
> ChromeHeadless has not captured in 60000 ms, killing.

Related to #307


karma.config.d/karma.conf.js
```diff
@@ -4,6 +4,10 @@ module.exports = function(config) {
         // default timeout 60_000, but looks like it is not enough for macOS on GitHub runner
         // increasing the timeout should help to reduce number of failures
         captureTimeout: 120_000,
+        // the default value is 20_000
```
krzema12, line 7: The error message mentions 60 s so I'm afraid it won't help, but let's try anyway!

krzema12 (approved):

PR #372, opened 2025-01-18
Reuse common resource access infra for YAML test suite

krzema12:
Thanks to this change, we reuse what was built to access common resources to also access YAML test suite data, allowing us to simplify the build logic and have a more uniform way of accessing both kinds of common resources.


src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/FSPath.kt
```diff
@@ -35,9 +36,21 @@ class FSPath(val path: Path, val fs: FileSystem) {
...
         FSPath(path.resolve(child), fs)
 
+    operator fun div(child: String): FSPath =
+        this.resolve(child)
+
     fun source(): Source =
         fs.source(path)
 
+    fun readUtf8(): String =
+        source().buffer().readUtf8()
```
OptimumCode, line 46: issue: the `readUtf8` method should called inside `use` block to make sure that resource is closed. It is not critical here since we are not opening a real resource but it still would be better to close resource correctly

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/FSPath.kt
```diff
@@ -35,9 +36,21 @@ class FSPath(val path: Path, val fs: FileSystem) {
...
+
+    /**
+     * Returns `null` if the file doesn't exist.
+     */
+    fun readUtf8OrNull(): String? =
+        if (this.exists) this.source().buffer().readUtf8() else null
```
OptimumCode, line 52: note: we could call newly created method `readUtf8` instead of duplicating the code here

OptimumCode (dismissed):

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/FSPath.kt
krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/372/commits/0bb98cac5aa21ca07e27049f71c33cbaa0789950

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/FSPath.kt
krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/372/commits/0bb98cac5aa21ca07e27049f71c33cbaa0789950

OptimumCode (approved):

PR #373, opened 2025-01-19
Allow reusing common test resources code generation

krzema12:
The goal is to allow accessing resources not only from `commonTest/resources`, but also from an arbitrary directory known during build time. In particular, we already generate code for the YAML comprehensive test suite ([buildSrc/src/main/kotlin/buildsrc/tasks/GenerateYamlTestSuiteData.kt](https://github.com/krzema12/snakeyaml-engine-kmp/blob/main/buildSrc/src/main/kotlin/buildsrc/tasks/GenerateYamlTestSuiteData.kt)), and we can replace it with the new code generation approach being adjusted in this PR. Thanks to this change, we'll be able to expose the YAML test data as a separate Okio file system, conveniently traversable in the tests.

This PR is just a piece of refactoring allowing it. Migrating the YAML test suite to use this approach will be done in a separate PR, to make things easier to review.

### DIFF
--- a/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/CommonTestResourcesAccess.kt
+++ b/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/CommonTestResourcesAccess.kt
@@ -33,12 +33,22 @@ fun stringFromResources(path: String): String {
 /**
  * Exposes files stored in `commonTest/resources` as a [FileSystem].
  */
-object CommonTestResourcesFileSystem : FileSystem() {
+val CommonTestResourcesFileSystem = buildFileSystem(CommonTestResources.resourcesMap)
+
+/**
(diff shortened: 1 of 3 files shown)


src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/CommonTestResourcesAccess.kt
```diff
@@ -33,12 +33,22 @@ fun stringFromResources(path: String): String {
 /**
  * Exposes files stored in `commonTest/resources` as a [FileSystem].
  */
-object CommonTestResourcesFileSystem : FileSystem() {
+val CommonTestResourcesFileSystem = buildFileSystem(CommonTestResources.resourcesMap)
```
krzema12, line 36: `CommonTestResourcesFileSystem` becomes just one (so far) of possible file systems created using the generated `resourceMap`.

OptimumCode (approved):

PR #375, opened 2025-01-20
Move code for JVM from 'java' to 'kotlin' directory

krzema12:
It's a purely cosmetic change, I think it doesn't influence anything functionally. However, we're heading towards the state of having 100% Kotlin code, and this is a Kotlin library, so it makes sense to cut off from Java wrt. directory structure.


krzema12:
@OptimumCode I was wrong about this change - green CI can also mean that some tests weren't run 😅 Here's a partial revert of this PR: https://github.com/krzema12/snakeyaml-engine-kmp/pull/376

PR #376, opened 2025-01-21
Move Java-based JVM tests back to 'java' directory

krzema12:
It's a follow-up after https://github.com/krzema12/snakeyaml-engine-kmp/pull/375. It's a partial revert of it - we can leave `jvmMain/kotlin` since there's no Java files in `jvmMain`.

I noticed this warning in the IDE:
Screenshot 2025-01-21 at 07 54 07

Also checked that the test isn't run.

Moving back to `java` fixed the problem.

Apparently:
* `java` directory can keep both Java and Kotlin code, to enable gradual migration
* `kotlin` directory can keep only Kotlin code

See https://kotlinlang.org/docs/gradle-configure-project.html#kotlin-and-java-sources:

> Do not store Java .java files in the src/*/kotlin directory, as the .java files will not be compiled.
> Instead, you can use src/main/java.


OptimumCode (commented):
question: why do we also move some of kotlin files back? They should be visible for java sources if we keep them in kotlin folder

OptimumCode (approved):

PR #384, opened 2025-01-25
Fix StringStreamDataWriter.toString to not consume buffer

krzema12:
During debugging, it was confusing to see some contents of the buffer when pausing in the debugger before an assertion, but then seeing the assertion fail. It didn't make sense why the assertion fails - the value of the buffer was as expected.

It turned out that the debugger called `toString()`, consuming the buffer, and leaving an empty buffer when performing the assertion. The solution here is to just `peek()` the buffer in `toString()`, not consume it.

### DIFF
--- a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/api/StringStreamDataWriter.kt
+++ b/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/api/StringStreamDataWriter.kt
@@ -16,5 +16,8 @@ internal class StringStreamDataWriter(
         buffer.writeUtf8(string = str, beginIndex = off, endIndex = off + len)
     }
 
-    override fun toString(): String = buffer.readUtf8()
+    /**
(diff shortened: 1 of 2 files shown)


src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/api/StringStreamDataWriter.kt
```diff
@@ -16,5 +16,8 @@ internal class StringStreamDataWriter(
         buffer.writeUtf8(string = str, beginIndex = off, endIndex = off + len)
     }
 
-    override fun toString(): String = buffer.readUtf8()
+    /**
+     * Returns the contents of the internal buffer, without altering the buffer's state.
+     */
+    override fun toString(): String = buffer.peek().readUtf8()
```
krzema12, line 22: TODO: add a test.

OptimumCode (approved):

PR #387, opened 2025-01-28
Kotlinize and commonize ScannerWithCommentEnabledTest

krzema12:
Part of https://github.com/krzema12/snakeyaml-engine-kmp/issues/269.


krzema12 (commented):

| Benchmark suite | Current: 9aa92f07d4df82eca79a8cbc955b296c0a58c874 | Previous: 36a1026737058d2c9c2d338a9029c6a22b1113bc | Ratio |
|-|-|-|-|
| `macosX64.DumpBenchmark.dumpMapWithAllTypes ( {} )` | `97.69855230805385` us/op | `60.455774551263985` us/op | `1.62` |

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/ScannerWithCommentEnabledTest.kt
```diff
@@ -0,0 +1,363 @@
...
+    val expectedValue = expectedIdIterator.next()
+    val id = token.tokenId
+    println("Expected value: $expectedValue")
+    println("Actual   value: $id")
+    println()
+    expectedValue shouldBeSameInstanceAs id
+
+    if (expectedScalarValueIterator != null && id == ID.Scalar) {
+        expectedScalarValueIterator.next() shouldBe (token as ScalarToken).value
+    }
```
OptimumCode, line 362: suggestion: I think we could replace calling `println` with call to `withClue` method

```kotlin
withClue(
    {
        buildString {
            appendLine("Expected value: $expectedValue")
            appendLine("Actual value: $id")
        }
    }
) {
    expectedValue shouldBeSameInstanceAs id
    if (expectedScalarValueIterator != null && id == ID.Scalar) {
        expectedScalarValueIterator.next() shouldBe (token as ScalarToken).value
    }
}
```

It looks a bit ugly but the main difference would be that this clue will appear in the error message in case of test failure (the `println` results will be visible only in the test's stdout)

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/ScannerWithCommentEnabledTest.kt
```diff
@@ -0,0 +1,363 @@
...
+        )
+    }
+
+    val token = sut.peekToken()
+    assertTokenEquals(
+        expectedIdIterator = expectedIterator,
+        expectedScalarValueIterator = expectedScalarValueIterator,
+        token = token,
+    )
+    expectedIterator.hasNext().shouldBeFalse()
```
OptimumCode, line 344: suggestion: I would add a clue here saying `unexpected tokens` or something like that. Right now we will only see `expected: false, actual: true`

OptimumCode (dismissed):

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/ScannerWithCommentEnabledTest.kt
krzema12: Oh, this was supposed to be used only in debugging, I forgot to remove it! I think we don't need the clue here, to not obscure the test.

OptimumCode (approved):

PR #389, opened 2025-01-28
Implement partial source reading in StreamReader

### LINKED ISSUE #388: Do not read the whole source in memory in StreamReader
The current implementation of `StreamReader` uses the `BufferedSource.readUtf8(): String` method to pull codepoints from the source. This method reads whole data from `BufferedSource` and builds a `String` that is later used to populate the codepoints array. This behavior can cause potential issues when reading large files (e.g. OutOfMemoryError for JVM target). It also might impact the performance (because we had to allocate large arrays - #266).

There is a `BufferedSource.readUtf8(byteCount: Long): String` method that can read a string with the provided limit in bytes. However, this will only work for ASCII codepoints. When the `BufferedSource` contains codepoints other than ASCII we might read only a part of the codepoint (because in UTF-8 codepoint can be encoded as [1,2,3 or 4 bytes](https://www.rfc-editor.org/rfc/rfc3629.html#section-3)). Because of that, we cannot use this method without some additional logic implemented (initially on our side but later it might be contributed to the Okio library).

I suggest we implement something similar to what is done in the original SnameYAML library (except using the `char[]` buffer to read data from the source - we don't have such an option in KMP):

- implement a method to read a **valid** UTF-8 string with at most `n` bytes in it
- add a parameter to configure that `n`. By default, it will be 1024

@krzema12 Could you please take a look at the above? Does it make sense to you? I already prepared the part for `implement a method` and will submit it as a PR shortly

OptimumCode:
We implement an extension method that uses underlying `Buffer` in `BufferedSource` to check the boundaries for that valid codepoint in the source

Closes #388


krzema12 (commented):

| Benchmark suite | Current: dbb08dd7f424d954b38c3543ceeb9bcc349a8dec | Previous: 36a1026737058d2c9c2d338a9029c6a22b1113bc | Ratio |
|-|-|-|-|
| `macosArm64.LoadingTimeBenchmark.loadsOpenAiSchema ( {"openAiYamlPath":"data/issues/kmp-issue-204-OpenAI-API.yaml"} )` | `60.927139820253174` ms/op | `33.95186695075758` ms/op | `1.79` |

OptimumCode:
> # ⚠️ **Performance Alert** ⚠️

Will rerun the benchmark after build is complete to make sure it is just a flaky run

OptimumCode:
> [#389 (review)](https://github.com/krzema12/snakeyaml-engine-kmp/pull/389#pullrequestreview-2579053019)
> 
> Will rerun the benchmark after build is complete to make sure it is just a flaky run

After re-running the results are better. Don't think the difference is related to the change

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/utils/BufferedSourceExtensions.kt
```diff
@@ -0,0 +1,110 @@
...
+    return buffer
+}
+
+@Suppress("NOTHING_TO_INLINE")
+// based on benchmarks it looks like having inline here actually makes a small difference in the performance
+private inline fun BufferedSource.sizeOfFullValidUtf8String(limitBytes: Long): Long {
+    val readBytes = limitBytes
+    val hasAll = request(readBytes)
+
+    val originalSize = if (hasAll) readBytes else buffer.size
```
krzema12, line 22: nit: probably no need to introduce another val:
```suggestion
    val hasAll = request(limitBytes)

    val originalSize = if (hasAll) limitBytes else buffer.size
```

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/utils/BufferedSourceExtensions.kt
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * @return the number of bytes that must follow [byte] to form a well-formed UTF8 codepoint
+ */
+@Suppress("NOTHING_TO_INLINE")
+private inline fun continuationBytesCountFor(byte: Byte): Int {
+    val asInt = byte.toInt()
+
+    return when {
+        // 110xxxxx
+        asInt and 0xe0 == 0xc0 -> 1
```
krzema12, line 100: nit: here and in other places where we use bitmasks: using binary literals would let us remove the comments and increase readability, here for example:

```suggestion
        asInt and 0b1110_0000 == 0x1100_0000 -> 1
```

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/utils/BufferedSourceExtensionsTest.kt
krzema12, line 1: Kudos for the comprehensive tests! 🙇

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/ReaderStringTest.kt
```diff
@@ -107,4 +112,64 @@ class ReaderStringTest : FunSpec({
         reader.peek(1) shouldBe 's'.code
         reader.peek(2) shouldBe 't'.code
     }
+
+    test("read first codepoint from a yaml that does not fit in memory") {
+        LargeSource(maxSizeBytes = 50L * 1024 * 1024 * 1024 /*50 GB*/).use { source ->
```
krzema12, line 117: nit: how about defining
```
private val Int.GiB get() = this.toLong() * 1024 * 1024 * 1024
```
then we can get rid of the comment?

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt
```diff
@@ -190,7 +191,7 @@ class StreamReader(
 
     private fun update() {
         try {
-            val buffer = stream.readUtf8()
+            val buffer = stream.readUtf8WithLimit(1024L)
```
krzema12, line 194: Let's discuss the value of `1024L`. I presume it's a trade-off: lower values would mean higher overhead in the internals of `readUtf8WithLimit`, and higher values would increase memory consumption. I'm wondering if increasing this value could lead to better performance. If you think alike, how about testing it on e.g. `1024L * 1024L`, and maybe some intermediate value?

Regardless of what value we'll figure out here, it's worth dropping a comment about this value.

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/utils/BufferedSourceExtensions.kt
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    // This can only happen if the edge bytes (n-3,n-2,n-1) are all continuation bytes.
+    // That would mean the UTF8 string is ill-formed but, as before, this will be handled by Okio
+    return originalSize
+}
+
+@Suppress("NOTHING_TO_INLINE")
+private inline fun isContinuationByte(byte: Byte): Boolean {
+    val asInt = byte.toInt()
```
krzema12, line 87: nit: regarding casting to int, see this piece of Okio, they don't do casting: https://github.com/square/okio/blob/26970d8c67309601706f8c404bdd9ee7e7e6057e/okio/src/commonMain/kotlin/okio/Utf8.kt#L124-L128 - I've tried removing the casting but it doesn't compile... Interesting.

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/utils/BufferedSourceExtensions.kt
OptimumCode: Thanks for spotting. Just forgot to remove after refactoring I did

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt
OptimumCode: I was actually thinking about moving this into `LoadingSettings` (as it is in SnakeYAML).
But for sure, I can check how different sizes impact performance in our benchmarks

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/utils/BufferedSourceExtensions.kt
OptimumCode: They have an internal extension function `Byte.and(int: Int)` that does conversion to an Int. We don't have that many bit operations so I decided not to add similar function here

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt
OptimumCode: I did a quick the with `1024 * 1024` buffer size - does not look like it makes the benchmark work any faster.
I will do additional checks tomorrow and let you know the results

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt
```diff
@@ -39,12 +40,20 @@ class StreamReader(
...
     private val stream: BufferedSource = if (stream is BufferedSource) stream else stream.buffer()
 
     private val name: String = loadSettings.label
 
     private val useMarks: Boolean = loadSettings.useMarks
 
+    /**
+     * Size in bytes what will be used to request a new portion of data from [stream]
+     */
+    private val bufferReadSize: Long = loadSettings.bufferSize.toLong()
```
krzema12, line 55: For posterity: right now it's 1024: https://github.com/krzema12/snakeyaml-engine-kmp/blob/cc71a2c8906f6c81ea178fcb141cad60e26c2e64/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/api/LoadSettingsBuilder.kt#L42

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/scanner/StreamReader.kt
krzema12: Thanks, for me, relying on snakeyaml-engine's default is good enough, and anything more would mean optimizing how snakeyaml-engine does it which is fine, but perhaps a topic for a separate PR.

krzema12 (approved):
Great job! 🚀

PR #390, opened 2025-02-02
Use Core Schema by default as recommended by the YAML specification

Vampire:


Vampire:
The YAML specification in "10.3. Core Schema" says
> This is the recommended default [schema](https://yaml.org/spec/1.2.2/#recommended-schemas) that YAML [processor](https://yaml.org/spec/1.2.2/#processes-and-models) should use unless instructed otherwise.

But SYE uses the JSON schema by default, probably because in the beginning it did not support schemas at all and there was not able to handle the Core schema types. (https://bitbucket.org/snakeyaml/snakeyaml-engine/issues/38/core-schema-cannot-parse-integers-encoded)

This PR changes here the default to be the Core Schema, as recommended by the YAML specification.

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/api/LoadSettingsBuilder.kt
```diff
@@ -49,7 +49,7 @@ class LoadSettingsBuilder internal constructor() {
     private var useMarks: Boolean = true
     private var envConfig: EnvConfig? = null // no ENV substitution by default
     private var codePointLimit: Int = 3 * 1024 * 1024 // 3 MB
-    private var schema: Schema = JsonSchema()
+    private var schema: Schema = CoreSchema()
```
krzema12, line 52: (A general thread for this PR - creating attached to a line of code to be able to respond in the thread - GitHub's inconsistency...)

As far as my understanding about the schemas go, I agree that we should go with changing the default. What I would push for, though, is having this default schema defined ideally in a single place. I think such a refactoring should go first, and then this functional PR would ideally be a one-liner. This way we ensure proper consistency as well (as long as the refactoring PR is properly reviewed).

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/schema/Schema.kt
```diff
@@ -17,7 +17,7 @@ import it.krzeminski.snakeyaml.engine.kmp.api.ConstructNode
 import it.krzeminski.snakeyaml.engine.kmp.nodes.Tag
 import it.krzeminski.snakeyaml.engine.kmp.resolver.ScalarResolver
 
-internal val DEFAULT_SCHEMA = JsonSchema()
+internal val DEFAULT_SCHEMA = CoreSchema()
```
krzema12, line 21: I realized it would be good to consider how it affects the main consumer of the lib - kaml. I think it's worth at least running its CI/tests with the core schema. I saw that kaml does try to implement extra things on top of snakeyaml-engine-kmp, and I'm wondering if/what things can break. For clarity, I didn't review kaml's code in depth (yet).

krzema12: Also, I'm not very comfortable with a scenario where this change breaks kaml so that its tests don't catch it, and it gets auto-released like this. Of course then we can go the reactive path: there's a bug report against kaml, we analyze it and decide how kaml should behave, cover it with tests, repeat. However, being a bit proactive will at least show us what tests will fail there, if any, and make us prepared for a proper adjustment there.

Vampire: How should it affect kaml?
Kaml does not use the high-level API, but the low-level one.
And the only cases where it uses `DumpSettings` and `LoadSettings` are to give it to `StreamReader` and `Emitter` which do not care about the schema that is set.

Indeed a big question is, whether kaml should not start to use the high-level API and also let its users decide which schema to use, to get proper behavior. But for that this question should get answered: https://kotlinlang.slack.com/archives/C7A1U5PTM/p1738516397615949
Summarized, whether the behavior is correct for a KxS to try to somehow squeeze data into the DAO classes like automatically transforming a String in the data to a number in the DAO or also vice-versa.
If this is not the case but it should adhere to the types of the underlying data, it should most probably be refactored to use the high-level API and provide the users the ability to specify the schema they want to be used, but I don't know.

But either way, for this PR it is absolutely irrelevant as it does not touch anything that kaml is currently using.
And even if, then it should adapt to those breaking changes when updating. 🤷‍♂️

krzema12: Ok, I think you convinced me. This change is just the right thing to do, considering [the recommendation from the YAML specification](https://github.com/krzema12/snakeyaml-engine-kmp/pull/390#issuecomment-2629604856).

Two more things:
1. I'm wondering if bumping the major version is needed here - maybe the minor version is enough? Formally it's a breaking change, but the question is _how_ breaking 😄 Thoughts?
2. @OptimumCode @aSemy I just want to keep you in the loop here since you belong to the project - please ack this change, or chime in!

OptimumCode: I think, how breaking the change is will vary for every case... The Core schema is an extension of JSON schema if I get the specification correctly. So, as long as users did not leave any uncertainties in their YAML files they should be fine... (e.g. `value: 0x3A` - is it a string or a number? With JSON schema it is a string... `value: '0x3A'` - now we see that it is a string indeed)

I think the best what we can do is to explain the breaking change in the release notes and describe the options to mitigate those changes.

As for kaml, we can probably make a PR to provide an option to choose which schema to use. I don't really see what else can be done here...

krzema12: Things like `0x3A` or leading zeros - I see it extremely rarely in YAMLs. This change almost feels like a bug fix (so bumping patch version), but it's a bit more - I think bumping the minor version will be fair.

OptimumCode: > I think bumping the minor version will be fair

I agree. At least it highlights that the release has something more than just a bug fix and users probably should go and check the release notes

krzema12 (approved):

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/schema/Schema.kt
krzema12: Thanks @Vampire and @OptimumCode , merging!

PR #391, opened 2025-02-02
Increase regex readability

Vampire:


Vampire:
This PR makes all regex patterns that contain escapes raw String literals, so that you do not have to mentally double unescape to understand the pattern.

It also removes unnecessary start- and end-of-input anchors for regex that are anyway only used to match entire inputs.
If the anchors should stay, they should at least be changed to always be used, currently in most cases it was like `^foo|bar|baz$` instead of `^(?:foo|bar|baz)$`, so only matching start-of-input for `foo` and end-of-input for `baz`.

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/constructor/json/ConstructYamlBinary.kt
```diff
@@ -25,7 +25,7 @@ class ConstructYamlBinary : ConstructScalar() {
         // Ignore white spaces for base64 encoded scalar
         // TODO decodeBase64() doesn't seem to require removing whitespace - perhaps this can be removed?
         //      I'm leaving it in because I'm not confident about edge case test coverage for whitespaces.
-        val noWhiteSpaces = constructScalar(node).replace("\\s".toRegex(), "")
+        val noWhiteSpaces = constructScalar(node).replace("""\s""".toRegex(), "")
```
OptimumCode, line 28: @suggestion: We could extract this Regex into a constant to reuse between invocations (like it is done for most of other Regex patterns)

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/common/Anchor.kt
```diff
@@ -54,6 +54,6 @@ class Anchor(val value: String) {
             '*',
             '&',
         )
-        private val SPACES_PATTERN = Regex("\\s")
+        private val SPACES_PATTERN = Regex("""\s""")
```
OptimumCode, line 57: suggestion: I would leave such small patterns with escape characters - for me it adds more noise around in such cases. What do you think @Vampire @krzema12 ?

OptimumCode (commented):
Hi, @Vampire. Thank you very much for the contribution

> If the anchors should stay, they should at least be changed to always be used, currently in most cases it was like ^foo|bar|baz$ instead of ^(?:foo|bar|baz)$

I think it would be better to correct the patterns so the anchors are used correctly - just a precaution in case an incorrect method on Regex is used (that does not match the entire string against the pattern).

Vampire:
> I think it would be better to correct the patterns so the anchors are used correctly - just a precaution in case an incorrect method on Regex is used (that does not match the entire string against the pattern).

Sure, I hope I didn't break it now. :-D

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/constructor/json/ConstructYamlBinary.kt
```diff
@@ -25,7 +25,11 @@ class ConstructYamlBinary : ConstructScalar() {
         // Ignore white spaces for base64 encoded scalar
         // TODO decodeBase64() doesn't seem to require removing whitespace - perhaps this can be removed?
         //      I'm leaving it in because I'm not confident about edge case test coverage for whitespaces.
-        val noWhiteSpaces = constructScalar(node).replace("\\s".toRegex(), "")
+        val noWhiteSpaces = constructScalar(node).replace(SPACES_PATTERN, "")
         return noWhiteSpaces.decodeBase64()?.toByteArray() ?: byteArrayOf()
     }
+
+    companion object {
```
OptimumCode, line 32: suggestion: I think if we add `private` modifier for companion object we should not get any changes in the API file

Vampire: I just followed your request to make it like the other cases where it is done like that. :-D
I just made it a `private val` in the Kotlin file now, unless you really prefer a private companion object.

OptimumCode: I'm not requesting anything) I just left some suggestions - everyone is free to disagree with them or ignore them completely) If something is an issue that doesn't let the PR go forward, in my opinion, it will be clearly indicated.

Regarding the change itself: I don't think we want to change a public API just because of introducing a constant (that is what just happened). Because of that I suggest we try to make the companion object private (or maybe internal) to avoid changes in public API

Vampire: Requesting, suggesting, doesn't make too much difference in this context. :-D
For me a collaborator doing a PR review and suggesting changes is requesting them to be done. :-D

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/resolver/CoreScalarResolver.kt
```diff
@@ -43,17 +43,17 @@ class CoreScalarResolver : BaseScalarResolver(
          * Be aware that this regex will also match integers.
          */
         val FLOAT = Regex(
-            "^([-+]?(\\.[0-9]+|[0-9]+(\\.[0-9]*)?)([eE][-+]?[0-9]+)?)" + // float
-                    "|([-+]?\\.(?:inf|Inf|INF))" + // infinity
-                    "|(\\.(?:nan|NaN|NAN))$", // not a number
+            """^(?:([-+]?(\.[0-9]+|[0-9]+(\.[0-9]*)?)([eE][-+]?[0-9]+)?)""" + // float
```
OptimumCode, line 46: Just for reference (I did know about non-capturing groups before - every day brings something new😀):

> unlike capturing groups, it does not memorize the matched text, allowing for better performance and avoiding confusion when the pattern also contains useful capturing groups

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Regular_expressions/Non-capturing_group

OptimumCode (dismissed):
All looks good to me, thank you @Vampire. @krzema12 have you had a chance to take a look at the PR? Would you have any suggestions/objections?

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/resolver/CoreScalarResolver.kt
Vampire: As a great resource for regex I can also heavily recommend https://www.regular-expressions.info/, especially when it is also about different dialects. :-)

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/resolver/CoreScalarResolver.kt
krzema12, line 1: It would be good to port https://github.com/krzema12/snakeyaml-engine-kmp/blob/e0ce6a42672ca547a6329f9fb3b5852919fe8e37/src/jvmTest/java/org/snakeyaml/engine/v2/resolver/CoreScalarResolverTest.java first.

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/resolver/JsonScalarResolver.kt
krzema12, line 1: It would be good to port [JsonScalarResolverTest.java](https://github.com/krzema12/snakeyaml-engine-kmp/blob/e0ce6a42672ca547a6329f9fb3b5852919fe8e37/src/jvmTest/java/org/snakeyaml/engine/v2/resolver/JsonScalarResolverTest.java) to KMP first.

Vampire:
Did you mistype and mean you **are** worried?
Because you said you are not.
Besides that, there were already non-capturing groups used in other regexes.
They are supported in JVM and JS.
On Native and WASM I'm not sure which flavor is used as the docs just say "this can be altered in the future to match the JVM behavior" so assuming it should be fine there too.

Vampire:
From a quick look I'd say Native and WASM is an own Regex implementation also supporting non-capturing groups.
And I guess the intention is to be consistent with JVM implementation and if not fix that, hence the hint that it might change into that direction.

krzema12:
> Did you mistype and mean you **are** worried? Because you said you are not.

Yes,sorry!

Yeah, things should probably work fine, although as a rule of thumb, if we touch some code, I'd prefer to have coverage for all targets. I can migrate the two tests tomorrow (they're pretty simple), and this way we'll increase confidence regarding correctness of this change.

src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/resolver/CoreScalarResolver.kt
krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/397

krzema12: I'm having some problems with porting: https://github.com/krzema12/snakeyaml-engine-kmp/pull/397/files#r1944128242

krzema12 (approved):

PR #392, opened 2025-02-03
Verify emitted events also for error test suite tests

Vampire:


Vampire:
Emitted events are now also validated for error test-suite tests.
This brought up another 32 test cases that do not behave like specified in the test suite. :-(
There is also now as deviations list for "result is correct but events were wrong" and the other two deviation lists were combined to "result is wrong".
It was quite confusing that for example `JEF9:02` and `L24T:01` were in the list "should fail but pass" while the test suite says they should pass.
This was the case because the emitted events were not correct and this was probably hacked in.
Now those are in the new deviations list and the the status is taken from the `data`, not from which list the id is added to.

krzema12 (approved):
Good stuff!

It changes the division of deviations (from incorrectly failing and incorrectly succeeding to failing+succeeding with expected events and failing+succeeding with unexpected events), but I think it's fine.

PR #398, opened 2025-02-04
Synchronize kotest version to 6.0.0.M1 via version catalog

krzema12:
I see that Renovate managed to bump two packages, and for other packages, CI is failing. Not sure if it's because the artifacts are broken, or the changes need to go together. To solve it, let's make Renovate bump versions of all kotest-related packages at once, for the sake of consistency.

This change downgrades some kotest-related libraries to 6.0.0.M1.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -37,9 +37,9 @@ kotlin {
             )
 
             dependencies {
-                implementation("io.kotest:kotest-framework-engine:6.0.0.M1")
-                implementation("io.kotest:kotest-framework-api:6.0.0.M1")
-                implementation("io.kotest:kotest-assertions-core:6.0.0.M1")
+                implementation(libs.kotest.framework.engine)
+                implementation(libs.kotest.framework.api)
+                implementation(libs.kotest.assertions.core)
                 implementation("org.jetbrains:annotations:26.0.2")
                 // Overridig coroutines' version to solve a problem with WASM JS tests.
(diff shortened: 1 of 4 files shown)


build.gradle.kts
```diff
@@ -37,9 +37,9 @@ kotlin {
             )
 
             dependencies {
-                implementation("io.kotest:kotest-framework-engine:6.0.0.M1")
-                implementation("io.kotest:kotest-framework-api:6.0.0.M1")
-                implementation("io.kotest:kotest-assertions-core:6.0.0.M1")
+                implementation(libs.kotest.framework.engine)
+                implementation(libs.kotest.framework.api)
+                implementation(libs.kotest.assertions.core)
                 implementation("org.jetbrains:annotations:26.0.2")
```
krzema12, line 43: We probably should use the catalog for all our dependencies, which can be done as a follow-up. For now, I'm addressing the problems related to kotest.

OptimumCode: There is a Gradle plugin that could help make the migration to version catalog faster 

https://github.com/littlerobots/version-catalog-update-plugin

But I have not tried it so not sure if it works okay with KMP projects

OptimumCode (approved):

build.gradle.kts
krzema12: I think that for a small project with little dependencies, it even has some value to go one by one to understand what is used and how, especially grouped dependencies. Thanks for sharing, though!

PR #407, opened 2025-02-12
Centralize definition of default schema

Vampire:


src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/api/DumpSettingsBuilder.kt
```diff
@@ -54,7 +54,7 @@ class DumpSettingsBuilder internal constructor() {
     private var indentWithIndicator = false
     private var dumpComments = false
     private var isDereferenceAliases = false
-    private var schema: Schema = JsonSchema()
+    private var schema: Schema = DEFAULT_SCHEMA
```
krzema12, line 57: In line 319, change
```
    * default [it.krzeminski.snakeyaml.engine.kmp.schema.JsonSchema]. These 3 are available
```
to
```
     * [it.krzeminski.snakeyaml.engine.kmp.schema.DEFAULT_SCHEMA]. These 3 are available
```

krzema12: There may be more of these!

Vampire: Indeed there was also the one on LoadSettingsBuilder that stated that CoreSchema is the default :-D

Should really an `internal` property be used in the function docs?

krzema12: I'm fine with both using an internal property (still better than the current duplication and lying), or making the property public.

Vampire: Do you see in the generated docs what value that property has in one of the cases? Otherwise it might not be too helpful. Or does that property then also need doc? And if so, does it need to be public to be included in the rendering or is internal sufficient for that?

krzema12: Tested with Dokka.

With just mentioning the constant:
Screenshot 2025-02-12 at 10 03 58

clicking it:
Screenshot 2025-02-12 at 10 04 44

Clicking on "(source)" does open the right piece of code.

> Otherwise it might not be too helpful.

After 2-3 clicks, one gets to the desired value. I think it's better than introducing duplication that may desynchronize at some point.  I'm fine with adding a KDoc to `DEFAULT_SCHEMA` - it would save the consumer one click, at the cost of introducing minor, local duplication 😄 

Unfortunately, KDoc doesn't support something like `{@value }` in Javadoc: https://stackoverflow.com/a/57357666

Vampire: Which click would you save with KDoc?
image

krzema12: Oh, right! My brain somehow ignored it because it's just a type, not the value, but it's enough!

krzema12 (approved):

PR #485, opened 2025-07-31
Set minimum JDK version to 11

krzema12:
Last year we enabled support for JDK 8: https://github.com/krzema12/snakeyaml-engine-kmp/issues/202.

However:
* (the driver of this PR) the newest kotest supports JDK 11+ (https://github.com/kotest/kotest/blob/569fb2bfe1427d0d8879003f420ea3aff50c2793/documentation/docs/release_6.0.md?plain=1#L95)
* kaml targets JDK11 (https://github.com/charleskorn/kaml/blob/81a5c797057608fb0785d22c1ce29143fd3af9be/build.gradle.kts#L111, https://github.com/charleskorn/kaml/blob/81a5c797057608fb0785d22c1ce29143fd3af9be/build.gradle.kts#L132)
* version 11 has been EOL for over 3 years (https://endoflife.date/oracle-jdk)
* TODO: check if the motivation related to Bukkit (https://github.com/krzema12/snakeyaml-engine-kmp/issues/202) supporting Java 8 still holds

### DIFF
--- a/buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform.gradle.kts
+++ b/buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform.gradle.kts
@@ -114,7 +114,7 @@ kotlin {
 }
 
 //region Java versioning
-val minSupportedJavaVersion = JavaVersion.VERSION_1_8
+val minSupportedJavaVersion = JavaVersion.VERSION_11
 
 // use Java 21 to compile the project
 val javaCompiler = javaToolchains.compilerFor(21)


krzema12:
@4o4E from your perspective, would it be a breaking change? Can you share some docs on Bukkit? I cannot find info about JDK compatibility.

OptimumCode (approved):

4o4E:
> @4o4E from your perspective, would it be a breaking change? Can you share some docs on Bukkit? I cannot find info about JDK compatibility.

Hi, thanks for your continued work on this project!

I'd like to share some context regarding the proposed upgrade of the minimum Java version to 11.

In my case, I'm developing a plugin for the Minecraft Bukkit API, and I support a wide range of Minecraft server versions from 1.13 up to 1.21. Many popular versions (especially 1.13–1.16) are still widely used today and only support Java 8 due to limitations in the server runtime. In the Minecraft community, each game version is often treated as a separate product and may be maintained or used for years—similar to how Java itself supports long-term LTS versions like Java 8 and Java 11 in parallel.

Because of this, my plugin needs to maintain compatibility with Java 8. If snakeyaml-engine-kmp raises the minimum version to Java 11, I may be forced to stay on an older version of this library indefinitely, and I won't be able to benefit from future updates or bug fixes. That would be unfortunate.

Since the core functionality of this library doesn't seem to require Java 11 features, would you consider:

- Keeping Java 8 as the target compatibility level?

- Or offering a dual-build setup (e.g., Java 8 target for runtime, Java 11 for tests)?

- Or separating test dependencies like Kotest into profiles that only require Java 11 during testing?

This would help a broader range of users—including those in legacy or constrained environments—continue to benefit from this great project.

Thanks for considering!

aSemy:
> @krzema12 I cannot find info about JDK compatibility.

I don't know about Bukkit, but I assume it has the same Java requirements as Minecraft itself. Minecraft bumped the required Java version from 8 to 11 in version 1.17 (released in 2021)
 https://minecraft.fandom.com/wiki/Java_Edition_1.17

aSemy:
> @4o4E  Many popular versions (especially 1.13–1.16) are still widely used today and only support Java 8 due to limitations in the server runtime.

Are there any published statistics that can indicate how well used each version Minecraft is? Even if it's a rough sample it'd help. The stats could either be for servers (e.g. "10% of servers use 1.16") or users (e.g. "1% of players use 1.16 and below.")

aSemy:
> Are there any published statistics that can indicate how well used each version Minecraft is? Even if it's a rough sample it'd help. The stats could either be for servers (e.g. "10% of servers use 1.16") or users (e.g. "1% of players use 1.16 and below.")

I found some stats here https://bstats.org/global/bukkit

There are 3 reported Minecraft versions lower than v1.17

- 1.16 - 2.4%
-  1.8 - 3.3%
- 1.12 - 2.8%

And 2 reported Java versions lower than v11

- Java 8 - 4.9%
- Java 10 - 0.0%

What the stats don't tell us is how many players use each version. Maybe the <1.17 servers are all dead and have very few players? Or they are incredibly popular and have significantly more players than >1.17 servers?

krzema12:
I'm kinda convinced by what @Vampire wrote, that is: strive to keep the compatibility with Java 8, and just run the tests with Java 11. It's not only for Bukkit, but anyone that happens to be still on 8 for whatever reasons.

@4o4E do you feel like contributing? Once we see a PR that enables the mentioned approach, it would be easier to discuss. Maybe it's indeed the simplest approach @aSemy @OptimumCode?

4o4E:
> I'm kinda convinced by what @Vampire wrote, that is: strive to keep the compatibility with Java 8, and just run the tests with Java 11. It's not only for Bukkit, but anyone that happens to be still on 8 for whatever reasons.
> 
> @4o4E do you feel like contributing? Once we see a PR that enables the mentioned approach, it would be easier to discuss. Maybe it's indeed the simplest approach @aSemy @OptimumCode?

I found the following method to set the runtime of the test to 11

```kotlin
tasks.named("jvmTest", Test::class).configure {
    javaLauncher.set(
        javaToolchains.launcherFor {
            languageVersion.set(JavaLanguageVersion.of(11))
        }
    )
}
```

krzema12:
@4o4E I'm afraid it just tells Gradle to use Java 11 to run the tests (without this change, we use 21), instead of telling to compile production sources for tests for Java 11. I'd appreciate a working PR.

krzema12:
I'm experimenting with it in https://github.com/krzema12/snakeyaml-engine-kmp/pull/400.

Vampire:
You are currently setting complete option `"-Xjdk-release=${minSupportedJavaVersion.majorVersion}"` for all JVM targets. Just change that to 11 for the compile tasks that involve Kotest?

Vampire:
While I wonder, you set a 21 toolchain, don't you? So if compilation uses that toolchain already, shouldn't compilation work as it is and only the launcher toolchain is necessary?

krzema12:
> you set a 21 toolchain, don't you?

Yes, here: https://github.com/krzema12/snakeyaml-engine-kmp/blob/ca6c4b82156667a63c0186436ffac854f56b5df1/buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform.gradle.kts#L32

> So if compilation uses that toolchain already, shouldn't compilation work as it is and only the launcher toolchain is necessary?

The precise error I'm getting, without any Java version tweaks and with kotest 6.0.0.M9, is:
```
Cannot inline bytecode built with JVM target 11 into bytecode that is being built with JVM target 1.8. Specify proper '-jvm-target' option.
```

So I think 11 as the Java version we're launching the tests with is not enough. It looks to me like we need an alternative production code compilation task just used for tests that targets Java 11 so that it's compatible with kotest.

4o4E:
> but then got this error:
> 
> ```
> Execution failed for task ':compileTestKotlinJvm'.
> > Inconsistent JVM Target Compatibility Between Java and Kotlin Tasks
>     Inconsistent JVM Target Compatibility Between Java and Kotlin Tasks
>       Inconsistent JVM-target compatibility detected for tasks 'compileTestJava' (1.8) and 'compileTestKotlinJvm' (11).
> ```
> 
> I think I'm going to need more help with this one.

You can try setting up compileTestJava at the same time

i haven't tried the code below to see if it works
```kotlin
tasks.withType<Test>().configureEach {
    javaLauncher.set(
        javaToolchains.launcherFor {
            languageVersion.set(JavaLanguageVersion.of(11))
        }
    )
}
```

Vampire:
> So I think 11 as the Java version we're launching the tests with is not enough.

Ah, I see, that's probably right.

> I think I'm going to need more help with this one.

Well, do what it tells you, also change it for the Java test compilation task.

krzema12:
Ok, I've got something that works according to what I described above: https://github.com/krzema12/snakeyaml-engine-kmp/pull/489
I feel it will be considered hacky in the Gradle world, and there should be more Gradle-ish way to achieve it, but it at least works as intended. We can always improve as a follow-up.

Issue #453, opened 2025-06-22
[Port from snakeyaml-engine] Fix Inline comments in flow mapping context


krzema12:
Ports https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/4964e4fdd1266f93525f824984c037234dd23741.

PR #504, opened 2025-08-11
[Port from snakeyaml-engine] Fix Inline comments in flow mapping context

krzema12:
Fixes https://github.com/krzema12/snakeyaml-engine-kmp/issues/453.

Ports https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/4964e4fdd1266f93525f824984c037234dd23741.

### DIFF
--- a/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/DumpCommentInFlowStyleTest.kt
+++ b/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/DumpCommentInFlowStyleTest.kt
@@ -1,12 +1,9 @@
 package it.krzeminski.snakeyaml.engine.kmp.comments
 
-import io.kotest.assertions.throwables.shouldThrow
-import io.kotest.assertions.withClue
 import io.kotest.core.spec.style.FunSpec
 import io.kotest.matchers.collections.shouldBeEmpty
 import io.kotest.matchers.collections.shouldHaveSize
 import io.kotest.matchers.shouldBe
-import io.kotest.matchers.string.shouldContain
 import it.krzeminski.snakeyaml.engine.kmp.api.DumpSettings
 import it.krzeminski.snakeyaml.engine.kmp.api.LoadSettings
 import it.krzeminski.snakeyaml.engine.kmp.api.lowlevel.Compose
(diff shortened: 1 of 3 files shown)


src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/DumpCommentInFlowStyleTest.kt
krzema12, line 1: A note to `test("ignoring comments")` (I cannot add a comment to this line): even though this test is removed in https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/4964e4fdd1266f93525f824984c037234dd23741, I'd leave it because it does add some coverage on the behavior. I'm not sure why it was removed in the upstream commit. Perhaps because the test file deals with comments, and we have `setParseComments(false)`?

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/DumpCommentInFlowStyleTest.kt
```diff
@@ -31,7 +28,7 @@ class DumpCommentInFlowStyleTest : FunSpec({
 
     test("flow with comments") {
         val loader = Compose(LoadSettings.builder().setParseComments(true).build())
-        val content = "{ url: text # comment breaks it\n}"
+        val content = "{url: text # comment breaks it\n}"
```
OptimumCode, line 31: @krzema12 do you know why the space was removed? Looks like a redundant change (even if it was done in the original commit)

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/DumpCommentInFlowStyleTest.kt
OptimumCode: I always try to follow this rule: if the functionality is still there I try to keep the test. I also don't see a clear reason why the test was removed in the original commit. Let's keep it

OptimumCode (approved):

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/comments/DumpCommentInFlowStyleTest.kt
krzema12: Yeah, looks suspicious. I'm afraid I won't be able to chase it down myself, though...

OptimumCode: Should we create a ticket for this in such case? I will not be able to have a look at it myself any time soon either

krzema12: Sure: https://github.com/krzema12/snakeyaml-engine-kmp/issues/512

Issue #506, opened 2025-08-11
[Port from snakeyaml-engine] Add tests for dividing strings with spaces


krzema12:
To be ported (all commits related to [this issue](https://bitbucket.org/snakeyaml/snakeyaml-engine/issues/47)):
* https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/7eea790d14c986276f0095dd87694d5510cb34fd
* https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/061a977e9830bbd300e3a3c0cddf553a2ff4275c
* https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/9dfb6b6a1a45f4802f4ae3de5b22353cdc101724
* https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/e55f1a6957d5f422f62dd61260ca68a2ac9a01a8

PR #517, opened 2025-08-13
[Port from snakeyaml-engine] Add tests for dividing strings with spaces

krzema12:
Fixes https://github.com/krzema12/snakeyaml-engine-kmp/issues/506.

Ports multiple changes since individual commits make changes that make the tests fail, only the last commit makes the tests pass:
 * https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/7eea790d14c986276f0095dd87694d5510cb34fd
 * https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/061a977e9830bbd300e3a3c0cddf553a2ff4275c
 * https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/9dfb6b6a1a45f4802f4ae3de5b22353cdc101724
 * https://bitbucket.org/snakeyaml/snakeyaml-engine/commits/e55f1a6957d5f422f62dd61260ca68a2ac9a01a8

### DIFF
--- a/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/issues/issue47/SetWidthTest.kt
+++ b/src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/issues/issue47/SetWidthTest.kt
@@ -14,7 +14,7 @@ import it.krzeminski.snakeyaml.engine.kmp.events.*
 class SetWidthTest : FunSpec({
     val stringToSerialize = "arn:aws:iam::12345678901234567890:foobarbaz:testing:testing2:role/github-actions-role/\${{ github.token }}"
 
-    test("emit plain string and split") {
+    test("emitted plain text via Dump with limited width should be split") {
         val settings = DumpSettings.builder().setWidth(80).build() // Intentionally limited.
         val writer = StringStreamDataWriter()
         val dump = Dump(settings)
@@ -25,7 +25,7 @@ class SetWidthTest : FunSpec({
         yaml.trim() shouldBe expected
     }
 
-    test("emit plain and split") {
+    test("emitted plain text via Emitter with limited width should be split") {
         val settings = DumpSettings.builder().setWidth(80).build() // Intentionally limited.
         val writer = StringStreamDataWriter()
         with(Emitter(settings, writer)) {
@@ -40,7 +40,7 @@ class SetWidthTest : FunSpec({
         yaml shouldBe expected
     }
 
-    test("emit plain and no split") {
+    test("emitted plain text via Emitter with width longer than emitted text should not be split") {
         val settings = DumpSettings.builder().setWidth(180).build() // Intentionally limited.
         val writer = StringStreamDataWriter()
         with(Emitter(settings, writer)) {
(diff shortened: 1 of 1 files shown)


src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/issues/issue47/SetWidthTest.kt
```diff
@@ -0,0 +1,78 @@
...
+import it.krzeminski.snakeyaml.engine.kmp.api.Dump
+import it.krzeminski.snakeyaml.engine.kmp.api.DumpSettings
+import it.krzeminski.snakeyaml.engine.kmp.api.Load
+import it.krzeminski.snakeyaml.engine.kmp.api.LoadSettings
+import it.krzeminski.snakeyaml.engine.kmp.api.StringStreamDataWriter
+import it.krzeminski.snakeyaml.engine.kmp.common.ScalarStyle
+import it.krzeminski.snakeyaml.engine.kmp.emitter.Emitter
+import it.krzeminski.snakeyaml.engine.kmp.events.*
+
+class SetWidthTest : FunSpec({
```
OptimumCode, line 14: suggestion: I think we need to make test names a bit more explicit. Right now (at least for me) it is hard to understand what test does with looking though the whole test method. E.g: 

"emit plain text and split" -> "emitted plain text with limited width should not be split" (or something similar)

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/issues/issue47/SetWidthTest.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    test("emit plain and split") {
+        val settings = DumpSettings.builder().setWidth(80).build() // Intentionally limited.
+        val writer = StringStreamDataWriter()
+        with(Emitter(settings, writer)) {
+            emit(StreamStartEvent())
+            emit(DocumentStartEvent(false, null, emptyMap()))
+            emit(ScalarEvent(null, null, ImplicitTuple(true, true), stringToSerialize, ScalarStyle.PLAIN))
+            emit(DocumentEndEvent(false))
```
OptimumCode, line 35: note: It is hard to understand what these `null` and `false` parameters mean.

suggestion: we could either use names for these arguments or (I would prefer this option) we could extract creation of these event objects into methods (if I am not mistaken only ScalarStyle is being changed) and use meaningful name for those methods (e.g `scalarWithStyle(style)`)

OptimumCode (approved):

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/issues/issue47/SetWidthTest.kt
krzema12: Fair point as well, let's use named arguments: https://github.com/krzema12/snakeyaml-engine-kmp/pull/517/commits/f2e8d045a8bc63526b9bc4c25a902d965459865e

src/commonTest/kotlin/it/krzeminski/snakeyaml/engine/kmp/issues/issue47/SetWidthTest.kt
krzema12: I'm thinking about writing down some guidelines on porting stuff from snakeyaml-engine, I. e. what things we left untouched, and what we'd like to improve.

OptimumCode: Sounds like a good idea

PR #528, opened 2025-08-19
Add kotlin-js-store

krzema12:
Per [the docs](https://kotlinlang.org/docs/js-project-setup.html#version-locking-via-kotlin-js-store):

> To follow a [recommended practice](https://classic.yarnpkg.com/blog/2016/11/24/lockfiles-for-all/), commit kotlin-js-store and its contents to your version control system. It ensures that your application is being built with the exact same dependency tree on all machines


kotlin-js-store/wasm/yarn.lock
krzema12, line 1: Hmm, it should be in a different location: https://github.com/krzema12/snakeyaml-engine-kmp/blob/70a5d670cb7be6349b8548f8ffd80a8fe9ea11dd/buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform.gradle.kts#L164

krzema12: Someone got it working here: https://github.com/whyoleg/cryptography-kotlin/blob/1d0f98f97614b347a583f6469a80c8a46b860a13/build.gradle.kts#L32 but I don't see the `org.jetbrains.kotlin.gradle.targets.wasm` package.

PR #534, opened 2025-08-22
On MacOS in CI, just run tests for MacOS/iOS targets

krzema12:
The immediate reason for this change is not being able to update to kotest 6.0.0 and Kotlin 2.2.0.
See e.g. https://github.com/krzema12/snakeyaml-engine-kmp/actions/runs/17149519010/job/48656766305:

```
java.lang.IllegalStateException: Errors occurred during launch of browser for testing.
> Task :jsBrowserTest
- ChromeHeadless
Please make sure that you have installed browsers.
Or change it via
browser {
    testTask {
        useKarma {
            useFirefox()
            useChrome()
            useSafari()
        }
    }
}
```

It's weird because there's Chrome in the runner image: https://github.com/actions/runner-images/blob/9943ce2a33e95ba38691c0b26d5079a3216718e1/images/macos/macos-15-arm64-Readme.md?plain=1#L92. In general, MacOS workers are pretty weak, and make the builds longer. This change leads to not running browser tests on MacOS, which should be fine since we run them on Linux and Windows - this coverage is enough.

For Linux and Windows runners, I'd like us to refrain for these changes for now, just to make a minimal change that fixes the current problems.

This change is inspired by https://github.com/OptimumCode/json-schema-validator/blob/6558910afd62b22ed65183db4c08e5cbebc3b8fa/buildSrc/src/main/kotlin/convention.multiplatform-tests.gradle.kts#L7.

### DIFF
--- a/.github/workflows/build.main.kts
+++ b/.github/workflows/build.main.kts
@@ -58,9 +58,13 @@ workflow(
                     gradleVersion = "wrapper",
                 ),
             )
+            val target = when (jobRunner) {
+                MacOSLatest -> "macOsAllTest"
+                else -> "build"
+            }
             run(
-                name = "Build",
-                command = "./gradlew build --stacktrace",
+                name = "Build/test",
+                command = "./gradlew $target --stacktrace",
             )
         }
     }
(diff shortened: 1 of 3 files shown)


krzema12:
Let me merge it, will gladly accept feedback asynchronously.

PR #537, opened 2025-08-23
Migrate usecases/tags tests to Kotlin

krzema12:
Part of https://github.com/krzema12/snakeyaml-engine-kmp/issues/269.


build.gradle.kts
```diff
@@ -38,6 +38,7 @@ kotlin {
                 implementation(libs.kotest.framework.engine)
                 implementation(libs.kotest.assertions.core)
                 implementation(libs.kotest.assertions.table)
+                implementation(libs.kotlinx.datetime)
```
OptimumCode, line 41: issue: if this library is used only in tests we should use `testImplementation` configuration instead

OptimumCode (changes requested):

PR #593, opened 2025-10-16
Generate copy DSL using KSP

krzema12:
It's a follow-up to https://github.com/krzema12/snakeyaml-engine-kmp/pull/592.

The goal is to generate the boilerplate using a Kotlin Symbol Processing processor (almost like a compiler plugin). Thanks to this, we can test the processor's logic, and gain higher confidence that the library's `copy` DSL is implemented correctly.

The processor isn't aware of the YAML library, so potentially it can be extracted to a separate library one day.

### DIFF
--- a/copy-dsl-ksp-processor/src/jvmTest/kotlin/it/krzeminski/copydsl/impl/CopyDslSymbolProcessorTest.kt
+++ b/copy-dsl-ksp-processor/src/jvmTest/kotlin/it/krzeminski/copydsl/impl/CopyDslSymbolProcessorTest.kt
@@ -61,22 +61,14 @@ class CopyDslSymbolProcessorTest : FunSpec({
         val jvmClassFile = result.generatedFiles
             .first { it.extension == "class" }
             .parentFile
-        println("jvmClassFile: $jvmClassFile")
-        println("jvmClassFile.absoluteFile: ${jvmClassFile.absoluteFile}")
-        println("jvmClassFile.absolutePath: ${jvmClassFile.absolutePath}")
-        println("jvmClassFile.absoluteFile.invariantSeparatorsPath: ${jvmClassFile.absoluteFile.invariantSeparatorsPath}")
-        println("Files inside:")
-        jvmClassFile.walkTopDown().forEach { println("- $it") }
         val classpathSeparator = if (":" in jvmClassFile.absolutePath) ";" else ":"
-        val args = listOf(
-            "-cp",
-            listOf(jvmClassFile.absolutePath, compilation.kotlinStdLibJar!!.absolutePath).joinToString(classpathSeparator),
-            "MainKt",
-        )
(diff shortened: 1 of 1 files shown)


krzema12 (commented):

| Benchmark suite | Current: ac70ad8d8973bab39ecceb52a224e27b221e2eaf | Previous: 11d3d9ea01a94547f522c1409b620d095de09ffd | Ratio |
|-|-|-|-|
| `macosArm64.LoadingTimeBenchmark.loadsOpenAiSchema ( {"openAiYamlPath":"data/issues/kmp-issue-204-OpenAI-API.yaml"} )` | `68.30110169375` ms/op | `38.504916312056736` ms/op | `1.77` |

api/snakeyaml-engine-kmp.api
```diff
@@ -48,7 +47,64 @@ public final class it/krzeminski/snakeyaml/engine/kmp/api/DumpSettings {
 	public final fun isUseUnicodeEncoding ()Z
 }
 
-public abstract interface class it/krzeminski/snakeyaml/engine/kmp/api/DumpSettings$MutableDumpSettings {
+public final class it/krzeminski/snakeyaml/engine/kmp/api/DumpSettingsCopyDslKt {
```
krzema12: The API changed (formally) because now we have a top-level interface and an extension function instead of both being inside the annotated classes. Its required because KSP cannot modify existing classes, it can only produce new entities. Since we haven't released the library since the last big breaking change, there's no need to account for this technically breaking change. The intended use of the the `copy` function remains unchanged.

krzema12: See:

> Its required because KSP cannot modify existing classes, it can only produce new entities.

I don't know if it's possible to work around it in a sensible way. I think IDE will still suggest what extension functions are available for a given type, however I cannot check right now.

krzema12: Oh, compatibility with Java - yeah, I'd actually assume we don't need to care about it that much since it's a Kotlin-centric library, as opposed to snakeyaml-engine.

copy-dsl-ksp-processor/api/copy-dsl-ksp-processor.api
```diff
@@ -0,0 +1,15 @@
+public abstract interface annotation class it/krzeminski/copydsl/api/CopyDsl : java/lang/annotation/Annotation {
```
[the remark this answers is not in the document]
krzema12: You're right, I wanted to avoid making this PR more complex by adjusting the build logic. I'll try to remember to remove the API dump for this module.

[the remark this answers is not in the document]
krzema12: https://github.com/krzema12/snakeyaml-engine-kmp/pull/596

Issue #600, opened 2025-10-20
Build-time module `copy-dsl-ksp-processor` required in runtime


krzema12:
See https://github.com/charleskorn/kaml/actions/runs/18636590628/job/53128584995?pr=776:

 ```
What went wrong:
Could not determine the dependencies of task ':kotlinNpmInstall'.
> Failed to query the value of task ':kotlinNpmInstall' property 'packageJsonFiles'.
   > Could not resolve all dependencies for configuration ':jsNpmAggregated'.
      > Could not find it.krzeminski:copy-dsl-ksp-processor:4.0.0.
        Searched in the following locations:
          - https://repo.maven.apache.org/maven2/it/krzeminski/copy-dsl-ksp-processor/4.0.0/copy-dsl-ksp-processor-4.0.0.pom
        Required by:
            root project 'kaml' > it.krzeminski:snakeyaml-engine-kmp:4.0.0 > it.krzeminski:snakeyaml-engine-kmp-js:4.0.0
```

This module is by mistake required by consumers, and is not meant to be published to Maven Central. It's likely because there's `implementation` here:

https://github.com/krzema12/snakeyaml-engine-kmp/blob/eb27cf0528bbf73f929f8ac3a10ca20824f60bd1/build.gradle.kts#L41

PR #601, opened 2025-10-20
Remove dependency on unpublished KSP processor module

krzema12:
Fixes https://github.com/krzema12/snakeyaml-engine-kmp/issues/600.

Fixes an issue where version 4.0.0 of this lib cannot be consumed because of a missing dependency, a local module that isn't published to Maven Central. This change just removes the need of depending on the module. It's achieved by moving the KSP annotation to the library, and parmeterizing the processor with the annotation's FQN.

Why the kind of dependency isn't changed to `compileOnly`: because for some targets, Gradle requires the dependency to be added as `api`, defying the purpose of `compileOnly`. For e. g. the JVM, it would work as expected.

### DIFF
--- a/api/snakeyaml-engine-kmp.api
+++ b/api/snakeyaml-engine-kmp.api
@@ -844,9 +844,6 @@ public final class it/krzeminski/snakeyaml/engine/kmp/exceptions/YamlVersionExce
 	public fun <init> (Lit/krzeminski/snakeyaml/engine/kmp/common/SpecVersion;)V
 }
 
-public abstract interface annotation class it/krzeminski/snakeyaml/engine/kmp/internal/CopyDsl : java/lang/annotation/Annotation {
-}
-
 public final class it/krzeminski/snakeyaml/engine/kmp/nodes/AnchorNode : it/krzeminski/snakeyaml/engine/kmp/nodes/Node {
 	public fun <init> (Lit/krzeminski/snakeyaml/engine/kmp/nodes/Node;)V
 	public fun getNodeType ()Lit/krzeminski/snakeyaml/engine/kmp/nodes/NodeType;
--- a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/CopyDsl.kt
+++ b/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/internal/CopyDsl.kt
@@ -6,4 +6,4 @@ package it.krzeminski.snakeyaml.engine.kmp.internal
  */
 @Target(AnnotationTarget.CLASS)
 @Retention(AnnotationRetention.SOURCE)
-public annotation class CopyDsl
+internal annotation class CopyDsl


api/snakeyaml-engine-kmp.api
```diff
@@ -844,6 +844,9 @@ public final class it/krzeminski/snakeyaml/engine/kmp/exceptions/YamlVersionExce
 	public fun <init> (Lit/krzeminski/snakeyaml/engine/kmp/common/SpecVersion;)V
 }
 
+public abstract interface annotation class it/krzeminski/snakeyaml/engine/kmp/internal/CopyDsl : java/lang/annotation/Annotation {
+}
```
krzema12: It's public to be able to resolve it in the processor's unit tests. I added a docstring saying that it's visible only for testing. It's also not really usable by the users of the lib, they cannot do anything useful with this annotation as it's used only in build time (unless repurposed by the user).

krzema12 (commented):

| Benchmark suite | Current: 317e5bc18c2f6cd7aac9c20f9b545ff806ff1487 | Previous: eb27cf0528bbf73f929f8ac3a10ca20824f60bd1 | Ratio |
|-|-|-|-|
| `macosX64.LoadingTimeBenchmark.loadsOpenAiSchema ( {"openAiYamlPath":"data/issues/kmp-issue-204-OpenAI-API.yaml"} )` | `253.58610368095236` ms/op | `136.12609521764705` ms/op | `1.86` |
| `macosArm64.DumpBenchmark.dumpMapWithAllTypes ( {} )` | `33.22355801324703` us/op | `20.933656191745644` us/op | `1.59` |

buildSrc/src/main/kotlin/buildsrc/conventions/lang/kotlin-multiplatform.gradle.kts
```diff
@@ -46,8 +46,20 @@ kotlin {
...
+        browser {
+            testTask {
+                enabled = !System.getProperty("os.name").lowercase().contains("windows")
+            }
+        }
+        nodejs {
+            testTask {
+                enabled = !System.getProperty("os.name").lowercase().contains("windows")
+            }
+        }
```
krzema12: TODO: report.

build.gradle.kts
```diff
@@ -38,7 +38,6 @@ kotlin {
             dependencies {
                 implementation(libs.okio)
                 implementation(libs.urlencoder.lib)
-                implementation(project(":copy-dsl-ksp-processor"))
```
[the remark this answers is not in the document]
OptimumCode: question: can we use `compileOnly' configuration instead? I honestly don't remember if it works for kmp, but if it works we should not see it in the dependencies for any of the targets. We can change that in pom.xml generated for each target when published locally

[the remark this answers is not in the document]
OptimumCode: Ah, sorry, missed the last part of the description

api/snakeyaml-engine-kmp.api
krzema12: Ah, now it's clear:

> The `internal` visibility modifier means that the member is visible within the same module. More specifically, a module is a set of Kotlin files compiled together

So the tests need the annotation to be public because the annotation is compiled in a different compiler run than the code in the test (where we call the compiler manually). I'll move the annotation to the file passed to the compiler manually, to reproduce the case we have in the lib's code.

PR #602, opened 2025-10-20
Remove printing from tests

krzema12:
This is the only standard output printing happening unconditionally in the tests. It adds to already noisy test logs, and I think it's fine to remove.


krzema12:
Merging without review - it's a trivial, low-risk change.

PR #692, opened 2026-06-13
Port upstream bc68379 (Issue 68) - ParserImpl comments

krzema12:
Port upstream commit [bc68379](https://github.com/snakeyaml/snakeyaml-engine/commit/bc68379) (Issue 68) - adds comment handling in ParserImpl, including new inner classes for handling pending comments and node content. Also updates CommentAfterAliasTest to expect success instead of ParserException when parseComments=true.

### DIFF
--- a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/parser/ParserImpl.kt
+++ b/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/parser/ParserImpl.kt
@@ -1011,6 +1011,7 @@ class ParserImpl(
             return when {
                 indentlessSequence && scanner.checkToken(Token.ID.BlockEntry) -> {
                     endMark = scanner.peekToken().endMark
+                    states.addLast(nextState)
                     state = ParseIndentlessSequenceEntryKey()
                     SequenceStartEvent(anchor, tag, implicit, FlowStyle.BLOCK, startMark, endMark)
                 }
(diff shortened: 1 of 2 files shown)


krzema12:
## Review: Port upstream bc68379 (Issue 68) - ParserImpl comments

### Fidelity check ✓

The port faithfully adapts the upstream logic from `ParserImpl.java` to Kotlin:

1. **`hasNodeContent()`** ✓ — Logic identical to upstream, correctly checking for content tokens after comments.

2. **Comment handling in `parseNode()`** ✓ — The flow matches upstream:
   - Collects `CommentToken`s when anchor/tag is present
   - If content follows → `ParseNodeWithPendingComments` emits comments first, then delegates to `ParseNodeContent`
   - If no content (empty scalar) → creates scalar event, pops state, sets `ParseDocumentEndThenComments`
   - Uses `states.removeLast()` which is equivalent to upstream Java `states.pop()`

3. **`ParseNodeWithPendingComments`** ✓ — Correctly re-enters itself to emit comments one at a time, then delegates to `ParseNodeContent`.

4. **`ParseNodeContent`** ✓ — Handles all cases (BlockEntry, Scalar, FlowSequenceStart, FlowMappingStart, BlockSequenceStart, BlockMappingStart, empty scalar, error) identically to upstream. Uses Kotlin nullable `Mark` instead of `Optional<Mark>` — appropriate adaptation.

5. **`ParseDocumentEndThenComments`** ✓ — DocumentEnd → pending comments → ParseDocumentStart chain matches upstream exactly.

6. **Test changes** ✓ — Removes the `if (parseComments)` / `shouldThrow<ParserException>` branches, matching upstream's fix. The upstream tests always use `parseComments = true`; the KMP version runs both `true`/`false`, which is fine.

### Suggestion 💡

The upstream ([current master](https://github.com/snakeyaml/snakeyaml-engine/blob/master/src/test/java/org/snakeyaml/engine/issues/issue68/CommentAfterAliasTest.java)) includes 2 additional test cases that this port doesn't have:
- `testCommentAfterTag` — testing `!!str` tag with inline comment
- `testCommentAfterAnchorAndTag` — testing both anchor and tag with comment

Consider adding these for parity, but not blocking.

Issue #684, opened 2026-05-25
[Design] Syncing with upstream 2026-05-25


krzema12:
## Last analyzed commit

```
be40e57ffc1b5bac2d485d77c1af2da7e35adc7d
```

## Upstream commits since then (graph)

```
* 5503e17 Issue 80: Fix L24T test in test suite
* 8838fbf Issue 80: fix DK95 in test suite
* 4ce68cb Issue 80: fix test code for the test suite
* edae5eb Issue 80: update the test suite to data-2022-01-17
* 75442fd Add test for Issue 79: fix test
* f90c8cf Add test for Issue 79: float values may loose precision
*   37a9288 Merge pull request #3 from altro3/update-libs
|\  
| * 23a1bd9 Updating plugin versions. Updating lib versions.
|/  
* 655f721 Update changes
*   d6cc5db Merge branch 'master' into issue-68
|\  
| *   68d31df Merge branch 'master' into issue-75
| |\  
| | * 8a30b31 Update changes
| | * cedf62b handle boundary condition related to surrogate chars
* | | bc68379 Fix issue 68 when comments after anchor/tag
|/ /  
* | c1d4d3c Allow 0x7F in quoted scalar
* | e63aa32 Merge branch 'master' into issue-75
|\| 
| *   9c726fa Merge remote-tracking branch 'origin/master' into issue-76-parse-json
| |\  
| | * 623230a Ignore CLAUDE.md
| * | 9810a28 Fix parsing of JSON with multiple consecutive TABs
| |/  
| * 5f7eff2 Update version for JDK 11
| * 0ed8f4f Update JUnit
| * c463a58 Update maven-resources-plugin
| * 93608b6 Update some Maven plugins
* | 2c85c8c Allow 7F
|/  
* 06b1c92 Add a test for issue 75
* 2070eb4 [maven-release-plugin] prepare for next development iteration
* db85065 [maven-release-plugin] prepare release snakeyaml-engine-3.0.1
* 243f992 Prepare 3.0.1
* 37a144f Update changes
* e8c75a9 Issue 74: Refactor packages to run test with module-info.java
* 0280e0e Add module-info.java
*   33680c3 Merge remote-tracking branch 'origin/master'
|\  
| * 636de54 [maven-release-plugin] prepare for next development iteration
| * 8bebbda [maven-release-plugin] prepare release snakeyaml-engine-3.0
| * c2eb39a Update changes
[...]

PR #693, opened 2026-06-13
Port upstream 8876c5c (Issue 72) - allowNonScalarKeys setting

krzema12:
Port upstream commit [8876c5c](https://github.com/snakeyaml/snakeyaml-engine/commit/8876c5c) (Issue 72) - adds an allowNonScalarKeys setting to LoadSettings and enforces it in Composer.composeMappingChildren. Includes a test for the new behavior.

**⚠️ Intentional divergence from upstream:** The upstream default is `false` (security-conscious), but this KMP port defaults to `true` for backward compatibility with existing YAML documents that use non-scalar mapping keys. Users coming from upstream should explicitly set `allowNonScalarKeys = false` if they want the safer default.

### DIFF
--- a/api/snakeyaml-engine-kmp.api
+++ b/api/snakeyaml-engine-kmp.api
@@ -72,9 +72,10 @@ public final class it/krzeminski/snakeyaml/engine/kmp/api/Load {
 
 public final class it/krzeminski/snakeyaml/engine/kmp/api/LoadSettings {
 	public fun <init> ()V
-	public fun <init> (Ljava/lang/String;Ljava/util/Map;Lit/krzeminski/snakeyaml/engine/kmp/api/LoadSettings$CollectionProvider;Lit/krzeminski/snakeyaml/engine/kmp/api/LoadSettings$CollectionProvider;Lit/krzeminski/snakeyaml/engine/kmp/api/LoadSettings$CollectionProvider;Lit/krzeminski/snakeyaml/engine/kmp/api/LoadSettings$SpecVersionMutator;IZZIZLjava/util/Map;Lit/krzeminski/snakeyaml/engine/kmp/env/EnvConfig;ZILit/krzeminski/snakeyaml/engine/kmp/schema/Schema;)V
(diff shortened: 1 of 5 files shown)


krzema12:
## Review: Port upstream 8876c5c (Issue 72) - allowNonScalarKeys setting

### Fidelity check

1. **`LoadSettings.kt`** — Field definition with KDoc matches upstream Java Javadoc. Default value matches the upstream builder's default (`false`).

   **Wait** — actually the default here is `true`, not `false`. Upstream sets `allowNonScalarKeys = false` in `LoadSettingsBuilder.java`. The PR description says "defaults to true for backward compatibility" which is a deliberate divergence. This is a meaningful behavioral difference: all existing tests in upstream that use non-scalar keys had to be updated to pass `allowNonScalarKeys = true`, while KMP does not need those updates because the default is already `true`.

2. **`Composer.kt`** ✓ — The guard at `composeMappingChildren` matches upstream exactly: checks `itemKey.nodeType != NodeType.SCALAR && !settings.allowNonScalarKeys`.

3. **`ComposerTest.kt`** ✓ — Test case matches upstream logic. Uses `LoadSettings(allowNonScalarKeys = false)` which is equivalent to upstream's `LoadSettings.builder().build()`.

4. **API dump update** ✓ — Correctly reflects the new getter/setter.

### Important: Default value divergence ⚠️

This is the only PR in this batch where the port intentionally diverges from upstream behavior. Upstream defaults to `false` (security-conscious), KMP defaults to `true` (backward-compatible). This should be clearly documented, as users coming from upstream might expect the safer default. Consider adding a note in the changelog or migration docs.

### Missing: No updates to existing tests that use non-scalar keys

Upstream had to update ~8 test files to explicitly enable `allowNonScalarKeys`. The KMP version doesn't need to because of the `true` default, but it's worth confirming the KMP equivalent tests exist and pass (CI is green).

krzema12:
**Follow-up:** I want to be more specific about the default value concern. Upstream commit 8876c5c intentionally made `allowNonScalarKeys = false` the default — this is a **security hardening change** to prevent potential issues with untrusted YAML sources (Issue 72). The KMP port defaults to `true` for backward compatibility, which silently undoes that hardening.

Recommendation: consider matching upstream's `false` default. This would:
- Require updating the same set of KMP test files that upstream updated (FuzzyCollectionTest, RecursiveMapTest, RecursiveSetTest, ReferencesTest, DumpWithoutSpaceTest, EmitterTest, ComposeSuiteTest) to explicitly pass `allowNonScalarKeys = true`
- Break backward compatibility for users relying on non-scalar keys without explicit opt-in

Alternatively, keep `true` as a deliberate KMP choice but make sure it's clearly documented in release notes as a divergence from upstream.

krzema12:
Acknowledged. The per-plan decision from issue #684 is to default to `true` for backward compatibility in KMP. The PR description now documents this divergence with a warning. If the preference changes, we can revisit flipping the default and updating the affected tests.

PR #695, opened 2026-06-13
Port upstream 9a01a3c - remove dead is ByteArray branch from representPrimitiveArray

krzema12:
Port upstream commit [9a01a3c](https://github.com/snakeyaml/snakeyaml-engine/commit/9a01a3c) - removes the dead is ByteArray branch from CommonRepresenter.representPrimitiveArray. In the KMP version, ByteArray is handled separately by representByteArray, so the is ByteArray branch in the when block is unreachable.

### DIFF
--- a/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/representer/CommonRepresenter.kt
+++ b/src/commonMain/kotlin/it/krzeminski/snakeyaml/engine/kmp/representer/CommonRepresenter.kt
@@ -127,7 +127,6 @@ open class CommonRepresenter(
     private val representPrimitiveArray = RepresentToNode { data ->
         val style = settings.defaultFlowStyle
         val iterableData = when (data) {
-            is ByteArray    -> data.asIterable()
             is ShortArray   -> data.asIterable()
             is IntArray     -> data.asIterable()
             is LongArray    -> data.asIterable()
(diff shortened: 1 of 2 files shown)


krzema12:
## Review: Port upstream 9a01a3c - remove dead ByteArray branch

### Fidelity check ✓

**`CommonRepresenter.kt`** — Removes `is ByteArray -> data.asIterable()` from `representPrimitiveArray`.

This is a faithful port of upstream's removal of the `byte.class` branch from `StandardRepresenter.representPrimitiveArray`. The KMP code structure differs (Kotlin's `when` vs Java's `if-else` chain), but the effect is the same:

- `ByteArray::class` is registered separately at line 187 → `representByteArray`
- `representPrimitiveArray` no longer lists `ByteArray::class` in its `init` registrations (lines 190–196)
- Therefore the `is ByteArray` branch in `representPrimitiveArray` is genuinely dead code — removing it is correct

### No issues — ready to merge ✅

krzema12:
**Update:** I also noticed the upstream commit 9a01a3c additionally adds a `representIterator()` test in `StandardRepresenterTest.java` that is not ported here. The KMP version does register `representIterator` in `parentClassRepresenters`, but there is no corresponding test for it in the test suite. Consider adding one for full parity with upstream, though not blocking.

